PR #2, opened 2023-06-16
 add reproducible builds

### LINKED ISSUE #1: Setup reproducible builds for cryptopower
https://reproducible-builds.org/

It should be possible to do cryptographic verification of the binary generated for cryptopower. This would ensure that the source code has not been tampered with in any way after compiling the application.

dreacot:
Closes #1 

[Reproducible builds](https://reproducible-builds.org/) is a way to verify that no vulnerabilities or backdoors have been introduced during this compilation process. By promising identical results are always generated from a given source, this allows multiple third parties to come to a consensus on a “correct” result, highlighting any deviations as suspect and worthy of scrutiny.

To execute the script on Linux or Mac, simply execute `./reproducible_builds.sh linux-amd64-binary` in the root folder of your project


reproduciblebuilds/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+e.g `./reproducible_builds.sh linux-amd64-binary`
+
+### Verification
+
+You can then verify if the binaries are reproducible using a tool like [diffoscope](https://diffoscope.org/) or [reprotest](https://salsa.debian.org/reproducible-builds/reprotest)
+
+#### Using Diffoscope
+
+e.g `diffoscope binary1 binary2`
```
dmigwi, line 27: Shouldn't you specify a standard reference binary the users can use to compare against?

dmigwi (changes requested):
@dreacot Would you add more comments on how are where users can obtain the start reference binary to be used with diffoscope.

reproduciblebuilds/README.md
dreacot: it's not necessary, reproducible build isn't a beginner or novice task, choosing to do it means you have an understanding of what you are trying to achieve which is, comparing a release binary to one you've replicated using docker. those binaries would be available on the release page, maybe i can add a link to the release page just for ease, and also specify where the reproduced binary can be found

reproduciblebuilds/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+The 2 binaries should be the official cryptopower release binary which can be found on the [release page](https://github.com/crypto-power/cryptopower/releases) and the binary you reproduced which can be found in this directory
```
dmigwi, line 31: I think here you meant to discuss about the official reference binary since the other one is the user generated from his own source code.

dreacot: that's what the comment says

dmigwi: >The 2 binaries should be the official cryptopower release binary which can be found on the ...

unless I am getting it wrong, the statement above implies the opposite of when you just said.

dreacot: part one of the statement

> The 2 binaries should be the official cryptopower release binary which can be found on the [release page](https://github.com/crypto-power/cryptopower/releases)

part 2 of the statement

> and the binary you reproduced which can be found in this directory

reproduciblebuilds/Dockerfile
```diff
@@ -1,4 +1,9 @@
-FROM golang 
+FROM golang:1.20 as builder
```
crux25, line 1: Our official documentation says go 1.19, we should use same version as the official doc or update official doc to 1.20.

PR #16, opened 2023-07-04
delete gitlab ci

dreacot:


dmigwi:
This branch seems to be failing because the GOPATH is not detected in the workflow instance. Can you hardcode the GOPATH variable?

dmigwi:
And add it to PATH env variable

PR #28, opened 2023-07-17
Update README.md to add mobile screenshots

dreacot:


dmigwi (approved):
LGTM although having matching screenshots between the desktop and mobile would have been awesome.

Issue #38, opened 2023-07-19
Navigation bar hidden behind the application title bar.


crux25:
The application navigation bar is partially hidden behind the title bar.

To reproduce clone and build the current master (Commit ac178b7)

Platform: Fedora linux on Wayland and Gnome 43.6

dreacot:
@crux25 is this still reproducible

crux25:
> @crux25 is this still reproducible

![Screenshot from 2023-11-02 08-35-51](https://github.com/crypto-power/cryptopower/assets/4796738/1dcd27de-79f8-4d9b-bbe5-9aad63d12adc)
Still reproducible on the current master.

PR #50, opened 2023-08-21
Fix overlapped top navigation bar

dmigwi:
Closes https://github.com/crypto-power/cryptopower/issues/38
Before: 

image

After:

image

### DIFF
--- a/ui/page/root/main_page.go
+++ b/ui/page/root/main_page.go
@@ -621,6 +621,9 @@ func (mp *MainPage) layoutDesktop(gtx C) D {
 				Width:       cryptomaterial.MatchParent,
 				Height:      cryptomaterial.MatchParent,
 				Orientation: layout.Vertical,
+				Padding: layout.Inset{
+					Top: values.MarginPadding24,
+				},
 			}.Layout(gtx,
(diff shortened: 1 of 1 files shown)


dreacot:
in your after screenshot, i can see a little space after the top bar, the top starting point should be right immediately after the app window.

for me, however the gap is much wider

Screenshot 2023-08-21 at 5 12 27 PM

dmigwi:
How does it look on your end on master? Can you see that the top bar
overlap?

dreacot:
Screenshot 2023-08-21 at 7 02 39 PM

dmigwi:
Maybe its only on my machine that its acting up. I remember crux reported this a while back. we can hold on this until we identify why its works on your end but fails on mine.

I hope go 1.21 version update has no relation with this UI overlap

dreacot:
> I hope go 1.21 version update has no relation with this UI overlap

i just updated to 1.21, still shows fine for me

dmigwi:
Then this must be something unique to my system. I  will dig more to
understand the course

Sirmorrison:
The PR has deprecated UI and currently stalled. Closing as Staled.

PR #59, opened 2023-09-08
prevent address discovery from happening each time the app is launched

### LINKED ISSUE #55: [LTC]: when address discovery is completed on a restored wallet, the appropriate flag isn't set
when address discovery is completed on a restored wallet, the flag isn't set as completed, which causes address discovery to happen again when next the app starts up.

Screenshot 2023-09-08 at 1 05 27 PM

dreacot:
closes #55


dmigwi (approved):
LGTM. This ought to have been deleted along time ago

Issue #62, opened 2023-09-11
create base structure for new overview page


dreacot:
This work does not involve populating the overview page, the scope of this work is limited to;

1. creating the 3 tabs(Overview, wallets, Trade), the overview tab can be left empty
2. moving the current wallet list page into the wallets tab
3. moving the exchange section into the trade tab

![image](https://github.com/crypto-power/cryptopower/assets/25265396/a8891c7f-bbcf-4979-a7e1-eaf967f2d692)

PR #65, opened 2023-09-13
Add new Homepage layout

Sirmorrison:
1st part of the task for #62


dreacot:
Looking good!

The exchange button below doesn't currently work, it was working previously

Screenshot 2023-09-13 at 1 18 25 PM

Clicking on a wallet, and then returning, the new layout no longer perists

Screenshot 2023-09-13 at 1 19 14 PM

dreacot:
i see the update PR description, i appreciate the tasks breakdown, however working functionality on the wallets tab previously should be functional.

as follow up PRs could take quite some time

ui/cryptomaterial/image.go
```diff
@@ -20,35 +20,38 @@ func NewImage(src image.Image) *Image {
 	}
 }
 
+// reduced the image original scale of 1 by half to 0.5 fix blurry images
+// this in turn reduced the imahe layout size by half. Multiplying the
```
dreacot, line 24:
```suggestion
// this in turn reduced the image layout size by half. Multiplying the
```

ui/assets/decredicons/notification.png
dmigwi, line 1: I think this Icon should be replaced with two Icons. A full bell icon could be used to show all notifications have been read. And an Icon with some red/whatever color circle could be used to show pending notifications.
Below is how github does it to show pending notifications.
image

ui/page/root/home_page.go
```diff
@@ -0,0 +1,357 @@
...
+		// TODO: Use real values as these are dummy so lint will pass
+		hp.ParentNavigator().Display(settings.NewSettingsPage(hp.Load))
+	}
+
+	for hp.appLevelSettingsButton.Clicked() {
+		hp.ParentNavigator().Display(settings.NewSettingsPage(hp.Load))
+	}
+
+	for hp.hideBalanceButton.Clicked() {
+		// hp.isBalanceHidden = hp.WL.SelectedWallet.Wallet.ReadBoolConfigValueForKey(sharedW.HideBalanceConfigKey, false)
```
dmigwi, line 145: remove this commented code if its not needed.

ui/page/root/home_page.go
```diff
@@ -0,0 +1,357 @@
...
+
+// OnNavigatedFrom is called when the page is about to be removed from
+// the displayed window. This method should ideally be used to disable
+// features that are irrelevant when the page is NOT displayed.
+// NOTE: The page may be re-displayed on the app's window, in which case
+// OnNavigatedTo() will be called again. This method should not destroy UI
+// components unless they'll be recreated in the OnNavigatedTo() method.
+// Part of the load.Page interface.
+func (hp *HomePage) OnNavigatedFrom() {
+	// Also disappear all child pages.
```
dmigwi, line 183: This comment could use some fixing.

ui/page/root/home_page.go
```diff
@@ -0,0 +1,357 @@
...
+// displayed.
+// Part of the load.Page interface.
+func (hp *HomePage) HandleUserInteractions() {
+	if activeTab := hp.CurrentPage(); activeTab != nil {
+		activeTab.HandleUserInteractions()
+	}
+
+	var pg app.Page
+	switch hp.navigationTab.SelectedIndex() {
+	case 0:
```
dmigwi, line 112: Because the index order might change in future why don't you use the tab names?

ui/cryptomaterial/tab_nav.go
```diff
@@ -0,0 +1,85 @@
...
+	"gioui.org/layout"
+	"gioui.org/op/clip"
+	"gioui.org/op/paint"
+
+	"github.com/crypto-power/cryptopower/ui/values"
+)
+
+type TabNavigation struct {
+	list          *ClickableList
+	selectedIndex int
```
dmigwi, line 15: Add `selectedTabName string` here to. This will help reduce reliance on the index that could change if the tabs are swapped.

ui/page/root/home_page.go
```diff
@@ -0,0 +1,357 @@
...
+				return layout.Inset{Top: values.MarginPadding5}.Layout(gtx, func(gtx C) D {
+					return hp.hideBalanceButton.Layout(gtx, icon.Layout24dp)
+				})
+			}),
+		)
+	})
+}
+
+// TODO: use real values
+func (hp *HomePage) LayoutUSDBalance(gtx C) D {
```
dmigwi, line 305: IMO this feature shouldn't exist because of its reliance on the exchange rate that are not standard.

ui/page/root/overview_page.go
```diff
@@ -0,0 +1,103 @@
...
+
+func (op *OverviewPage) layoutDesktop(gtx C) D {
+	return op.comingSoonLayout(gtx)
+}
+
+func (op *OverviewPage) layoutMobile(gtx C) D {
+	return op.comingSoonLayout(gtx)
+}
+
+func (op *OverviewPage) comingSoonLayout(gtx C) D {
```
dmigwi, line 89: This is not a proper method name appearing in the code. This `"Overview Page Coming soon....."` should be enough to show page is under construction.

ui/page/root/trade_page.go
```diff
@@ -0,0 +1,103 @@
...
+
+func (tp *TradePage) layoutDesktop(gtx C) D {
+	return tp.comingSoonLayout(gtx)
+}
+
+func (tp *TradePage) layoutMobile(gtx C) D {
+	return tp.comingSoonLayout(gtx)
+}
+
+func (tp *TradePage) comingSoonLayout(gtx C) D {
```
dmigwi, line 89: This is not a proper method name appearing in the code. This `"Overview Page Coming soon....."` should be enough to show page is under construction.

ui/cryptomaterial/tab_nav.go
```diff
@@ -0,0 +1,85 @@
...
+
+	"gioui.org/layout"
+	"gioui.org/op/clip"
+	"gioui.org/op/paint"
+
+	"github.com/crypto-power/cryptopower/ui/values"
+)
+
+type TabNavigation struct {
+	list          *ClickableList
```
dmigwi, line 14: Why can't you reuse the existing component `NewTabItem`? Add your unique UI upgrades to that component.

dmigwi (changes requested):
The UI looks nice and well designed but something looks off. This UI shouldn't look like this.
image

IMO this feature should not be there because of it dependence on the exchange rates that are not standard. If we have to show the total value in respective coins, we can add it in the wallets lists 
Since the wallet name is restricted to 20 characters, the wallet balance could be added after it in 4-5 significant figures. This will consume a max of 10 characters. Wallet in total Name + Balance would consume a max of 30 characters.

ui/assets/decredicons/notification.png
dreacot: @Sirmorrison is following a UI guide provided by the designers, so design changes cannot be addressed in this PR

ui/page/root/home_page.go
```diff
@@ -0,0 +1,352 @@
...
+	switch hp.navigationTab.SelectedIndex() {
+	case 0:
+		pg = NewOverviewPage(hp.Load)
+	case 1:
+		pg = NewWalletDexServerSelector(hp.Load)
+	case 2:
+		pg = NewTradePage(hp.Load)
+	}
+
+	if pg != nil && hp.ID() != hp.CurrentPageID() {
```
itswisdomagain, line 119:
```suggestion
	if pg != nil && pg.ID() != hp.CurrentPageID() {
```

itswisdomagain (changes requested):

ui/cryptomaterial/tab_nav.go
Sirmorrison: The reason is the NewTabItem codebase seems deprecated and will need a major clean up job.
Adding code along side the deprecated codes, does not work well for me plus the lines of codes in that tabs.go file is already over 300.

ui/page/root/trade_page.go
Sirmorrison: I get your point however, this is just a temporary code and i didnt want to repeat the lines of codes in the Mobile and desktop methods.

This method will be removed almost immediately.

ui/page/root/overview_page.go
Sirmorrison: I get your point however, this is just a temporary code and i didnt want to repeat the lines of codes in the Mobile and desktop methods.

This method will be removed almost immediately.

ui/page/root/home_page.go
Sirmorrison: This is based on the current UI designs shared.

ui/assets/decredicons/notification.png
dmigwi: Ok. Noted. Where are the designs hosted at? Is it possible to pin the link somewhere they can easily accessed on demand i.e. the [project description on github](https://github.com/orgs/crypto-power/projects/4/settings)?

dmigwi:
@Sirmorrison The designs show some drop downs, is this coming in the next MR?

image

Sirmorrison:
> @Sirmorrison The designs show some drop downs, is this coming in the next MR?
> 
> image

Yes it is.. This PR is just for the UI layout.

dmigwi:
You've introduced some breaking changes. It would be best you handle the clean up in a separate PR.

image

Sirmorrison:
> You've introduced some breaking changes. It would be best you handle the clean up in a separate PR.
> 
> image

done

dreacot (approved):
Seen some bugs, should be fixed in a follow up PR, looks good overall

PR #108, opened 2023-09-27
add receive modal

### LINKED ISSUE #90: shows the corresponding page when the general app receive page is clicked
Screenshot 2023-09-25 at 2 44 55 AM

dreacot:
closes #90 

Screenshot from 2023-09-28 15-35-56

### DIFF
--- a/ui/page/components/receive_modal.go
+++ b/ui/page/components/receive_modal.go
@@ -38,13 +38,13 @@ type ReceiveModal struct {
 	sourceAccountSelector *WalletAndAccountSelector
 	sourceWalletSelector  *WalletAndAccountSelector
 
-	isNewAddr, isInfo        bool
-	currentAddress           string
-	qrImage                  *image.Image
-	newAddr                  cryptomaterial.Button
-	info, more               cryptomaterial.IconButton
-	receiveAddress           cryptomaterial.Label
-	generateNewAddressButton cryptomaterial.Button
+	isNewAddr, isInfo bool
+	currentAddress    string
+	qrImage           *image.Image
+	newAddr           cryptomaterial.Button
+	info, more        cryptomaterial.IconButton
+	receiveAddress    cryptomaterial.Label
+	// generateNewAddressButton cryptomaterial.Button
 }
 
 func NewReceiveModal(l *load.Load) *ReceiveModal {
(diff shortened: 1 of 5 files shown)


ui/page/components/receive_modal.go
```diff
@@ -0,0 +1,452 @@
...
+		rm.isNewAddr = !rm.isNewAddr
+		if rm.isInfo {
+			rm.isInfo = false
+		}
+	}
+
+	if rm.newAddr.Clicked() {
+		newAddr, err := rm.generateNewAddress()
+		if err != nil {
+			log.Debug("Error generating new address" + err.Error())
```
ukane-philemon, line 129: nit: needs space before the closing quotation mark.

ukane-philemon, line 129: nit: needs space before the closing quotation mark.

ui/page/components/receive_modal.go
```diff
@@ -0,0 +1,452 @@
...
+	var buffer bytes.Buffer
+	err = qrImage.SaveTo(&buffer)
+	if err != nil {
+		log.Error(err.Error())
+		return
+	}
+
+	decodeImg, _, err := image.Decode(bytes.NewReader(buffer.Bytes()))
+	if err != nil {
+		log.Error(err.Error())
```
ukane-philemon, line 183: Same as my comment above.

ui/page/components/receive_modal.go
```diff
@@ -0,0 +1,452 @@
...
+
+generateAddress:
+	newAddr, err := rm.sourceWalletSelector.selectedWallet.NextAddress(rm.sourceAccountSelector.selectedAccount.Number)
+	if err != nil {
+		return "", err
+	}
+
+	if newAddr == rm.currentAddress {
+		goto generateAddress
+	}
```
ukane-philemon, line 200: Is this possible?

ukane-philemon (changes requested):
Testing well except for a few touchups needed:

The copy toast alert does not sit well with the current position, I think. Maybe below the qr code? There's some estate there.

Screenshot 2023-09-28 at 11 00 50 PM

The Generate Address btn becomes transparent on hover, is that intended?

Also, I think clicking outside of the Generate Address btn should close the more display, no?
Screenshot 2023-09-28 at 11 00 26 PM

ui/page/components/receive_modal.go
dreacot: yes!, magic of go

ukane-philemon: Yh. I meant to confirm if this case `newAddr == rm.currentAddress` is ever possible, if not we can skip it.

dreacot: oh, my bad, yes there's a bug we discovered a while back, so this was a hack to bypass it, https://github.com/planetdecred/dcrlibwallet/pull/241/files the bug was fixed here but the code of the receive page was never updated, i ported over code from the receive page, i'd update it now, good catch

dreacot:
> The copy toast alert does not sit well with the current position, I think. Maybe below the qr code? There's some estate there.

This is displayed using `rm.Load.Toast.Notify(values.String(values.StrCopied))` there's no way to change the position except the Notify method is modified to receive a position variable, which i think would be overkill for this PR

ukane-philemon:
> `ify(values.String(values.StrCopied))` there's no way to change the position except the Notify method is modified to receive a position variable, which i think would be overkill for this PR

Okay.

ukane-philemon (approved):
Testing well.

PR #113, opened 2023-09-27
Fix wallet sync starting on app launch

### LINKED ISSUE #107: When the app is launched syncing doesn't start for wallets until you navigate to the wallets tab
Image

CPunch:
closes #107 

kind of wanted to give a shot at this, all of the wallet sync code is in wallet_selector_page.go, would it make sense to split it out or would a hacky fix like this be suitable?


crux25:
It is better to split the code, move the sync startup logic to home page. Calling wallet selector's navigation function while on home page will potentially introduce bugs in future, devs will assume that `OnNavigatedTo()` and  `OnNavigatedFrom()` are called only when the page is created and destroyed.

ui/page/root/home_page.go
```diff
@@ -102,6 +102,11 @@ func (hp *HomePage) OnNavigatedTo() {
 	hp.ctx, hp.ctxCancel = context.WithCancel(context.TODO())
 
 	if hp.CurrentPage() == nil {
+		// start wallet sync
+		pg := NewWalletSelectorPage(hp.Load)
+		pg.OnNavigatedTo()
+		pg.OnNavigatedFrom()
```
Sirmorrison, line 108: This is really hacky lol.

The only thought coming to mind is to load the wallets, and have them sync in background pending when we view the loaded wallets on the wallets page.

dreacot:
@CPunch please fix branch conflicts

ui/page/root/home_page.go
```diff
@@ -407,3 +450,104 @@ func (hp *HomePage) notificationSettingsLayout(gtx C) D {
...
+				// reset ticker
+				ticker.Reset(duration)
+			}
+			// Increase the counter
+			count++
+			log.Debugf("Attempting to check for internet connection in %s", duration.String())
+
+			go func() {
+				hp.isConnected = libutils.IsOnline()
+			}()
```
ukane-philemon, line 521: I don't think there’s a need for the nested goroutine.

ui/page/root/home_page.go
```diff
@@ -85,6 +92,21 @@ func NewHomePage(l *load.Load) *HomePage {
...
+	}()
+
+	// init shared page functions
+	toggleSync := func(unlock load.NeedUnlockRestore) {
+		if hp.WL.SelectedWallet.Wallet.IsConnectedToNetwork() {
+			go hp.WL.SelectedWallet.Wallet.CancelSync()
+			unlock(false)
+		} else {
+			hp.startSyncing(hp.WL.SelectedWallet.Wallet, unlock)
+		}
```
ukane-philemon, line 106: I think this check is inverted.

ukane-philemon (commented):
Quite straight forward.

ukane-philemon:
Looks good, you can update the title.

ukane-philemon (changes requested):
Almost.

dreacot:
@CPunch you have branch conflicts

ui/page/root/home_page.go
```diff
@@ -151,7 +151,6 @@ func (hp *HomePage) OnNavigatedTo() {
 		}
 	}
 
-	hp.CurrentPage().OnNavigatedTo()
```
CPunch: looks like this caused a crash else where. when this line is included, we get a panic at 

```
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x134e042]

goroutine 45 [running]:
github.com/crypto-power/cryptopower/ui/page/root.(*OverviewPage).listenForMixerNotifications.func1()
        /home/cpunch/projects/cryptopower/ui/page/root/overview_page.go:878 +0x2c2
created by github.com/crypto-power/cryptopower/ui/page/root.(*OverviewPage).listenForMixerNotifications in goroutine 37
        /home/cpunch/projects/cryptopower/ui/page/root/overview_page.go:848 +0x6f8
```

ui/page/root/home_page.go
```diff
@@ -405,6 +450,109 @@ func (hp *HomePage) notificationSettingsLayout(gtx C) D {
...
+	}
+	unlock(true)
+
+	if hp.isConnected {
+		// once network connection has been established proceed to
+		// start the wallet sync.
+		if err := wallet.SpvSync(); err != nil {
+			log.Debugf("Error starting sync: %v", err)
+		}
+		return
```
ukane-philemon, line 466: Missed this. I think we should fall through (i.e remove the `return`) so we get to try again after the first 10 sec until we successfully start syncing.

ukane-philemon (changes requested):
@CPunch, I think this PR is GTG after this.

ui/page/root/home_page.go
```diff
@@ -405,6 +449,109 @@ func (hp *HomePage) notificationSettingsLayout(gtx C) D {
 	)
 }
 
+func (hp *HomePage) startSyncing(wallet sharedW.Asset, unlock load.NeedUnlockRestore) {
+	// Watchonly wallets do not have any password neithers need one.
```
ukane-philemon, line 453: nit: neithers -> neither

ukane-philemon (approved):

dreacot (approved):

PR #141, opened 2023-10-03
Prevent duplicate notification registration

### LINKED ISSUE #109: UI: Error adding tx and block notification listener: listener_already_exist
Error observed when using the app for the first time.
```
[ERR] UI: Error adding tx and block notification listener: listener_already_exist
```
Sorry I don't have any other context at this time not sure exactly when it happened.

ukane-philemon:
closes #109.

To reproduce the issue: Navigate between the wallet_select page and the wallet main page in quick succession twice.

### DIFF
--- a/ui/page/info/info_page.go
+++ b/ui/page/info/info_page.go
@@ -2,6 +2,7 @@ package info
 
 import (
 	"context"
+	"sync/atomic"
 
 	"gioui.org/font"
 	"gioui.org/layout"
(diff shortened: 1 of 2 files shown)


ukane-philemon:
A wild goroutine is responsible for cleaning up and deregistering the listeners but might not have concluded before the request to register again.

I’ll revisit to reduce code changes.

ukane-philemon:
@dreacot I’m thinking about listeners having their separate I’d. This is because a wild goroutine can potentially deregister a listener for a displayed page with our current pattern.

Page adds listener > Quickly switching to and fro would cause same listener to be re-registered before the cleanup select case executes. We get an error about multiple listener registrations and then the wild goroutine eventually fires and clears up the first listener since they share a command unique id. Then we have no listener at all for the page.

dreacot:
> @dreacot I’m thinking about listeners having their separate I’d. This is because a wild goroutine can potentially deregister a listener for a displayed page with our current pattern.
> 
> Or do the cleanup when we cancel the page context.
> 
> Problem: Page adds listener > Quickly switching to and fro would cause same listener to be re-registered before the cleanup select case executes. We get an error about multiple listener registrations and then the wild goroutine eventually fires and clears up the first listener since they share a common unique id. Then we have no listener at all for the page.

go for it

ukane-philemon:
Replaced by #195

PR #147, opened 2023-10-04
Improve infinite scroll functionality

Sirmorrison:
It resolves issues with transaction data not being fetch for a filter even when filter should have transaction.
It also resolves issues with dropdown not showing up due to changes introduced in this commit https://github.com/crypto-power/cryptopower/commit/63514a3decef5c940376bfd69ef105bbd4d86ded#diff-4a73650fb46ca5164bb0be75cb4d0dc715b0b4f8b9b3966c4df1ba043c55fa7bR162

### DIFF
--- a/ui/cryptomaterial/dropdown.go
+++ b/ui/cryptomaterial/dropdown.go
@@ -1,6 +1,7 @@
 package cryptomaterial
 
 import (
+	// "fmt"
 	"image/color"
 
 	"gioui.org/layout"
@@ -284,8 +285,6 @@ func (d *DropDown) closedLayout(gtx C, iLeft int, iRight int) D {
 				layout.Rigid(func(gtx C) D {
 					return d.layoutOption(gtx, d.selectedIndex)
 				}))
-			w := (lay.Size.X * 800) / gtx.Dp(MaxWidth)
-			d.Width = w + 10
 			return lay
 		})
 	})
--- a/ui/page/components/items_scroll.go
+++ b/ui/page/components/items_scroll.go
@@ -15,7 +15,7 @@ import (
 // ScrollFunc is a query function that accepts offset and pagesize parameters and
 // returns data interface, count of the items in the data interface, isReset and an error.
 // isReset is used to reset the offset value.
-type ScrollFunc func(offset, pageSize int32) (data interface{}, count int, isReset bool, err error)
+type ScrollFunc func(offset, pageSize int32) (data interface{}, count int, err error)
 
 type Scroll struct {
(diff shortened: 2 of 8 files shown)


dreacot:
> if i time my scrolling well enough that the scroll bar start at the top on the next page, i'm unable to return to the previous page

i still have issues with the scrollbar

plus i think it's bad UX for the loader to show each time you are scrolling into a new item

[Screencast from 2023-10-04 11.18.34.webm](https://github.com/crypto-power/cryptopower/assets/25265396/67db93dc-ea9f-423b-a958-8022b941b2cb)

ui/cryptomaterial/dropdown.go
```diff
@@ -1,6 +1,7 @@
 package cryptomaterial
 
 import (
+	// "fmt"
```
dreacot, line 4: remove commented code

dreacot (changes requested):

dreacot:
on linux, the dropdown selector keeps increasing in width, until it takes full width

Screenshot from 2023-10-04 11-41-02

ui/cryptomaterial/dropdown.go
```diff
@@ -284,8 +285,6 @@ func (d *DropDown) closedLayout(gtx C, iLeft int, iRight int) D {
 				layout.Rigid(func(gtx C) D {
 					return d.layoutOption(gtx, d.selectedIndex)
 				}))
-			w := (lay.Size.X * 800) / gtx.Dp(MaxWidth)
-			d.Width = w + 10
```
ukane-philemon, line 288: I think removing this code caused this issue mentioned by @dreacot: https://github.com/crypto-power/cryptopower/pull/147#issuecomment-1747362071

I think this will fix the dropdown issue here. Consider replacing the deleted code with the suggested code below:  

```
if d.Width <= 0 {
    width := gtx.Dp(values.MarginPadding180)
    if d.revs {
	width = gtx.Dp(values.MarginPadding140)
    }
   d.Width = width
}
			
```

Also, this block of code is duplicated in `d.layoutOption(....)`. We can extract it into a `defaultDropdownWidth` method that returns the width based on the value of `d.revs`.

ukane-philemon (changes requested):

ui/cryptomaterial/dropdown.go
ukane-philemon: This fix has been moved to #188

Sirmorrison:
There is a follow up PR so this can be closed

PR #148, opened 2023-10-04
fix exchange rate bugs

### LINKED ISSUE #57: the binance rate seems off when creating an order
Screenshot 2023-09-08 at 4 22 30 PM

ukane-philemon:
Closes #57 

- Return early for `dcr-ltc` market as Binance does not support it yet.
- Fix the bug that caused weird estimated rates.
- Improve the estimated rates display so it's clear.
- Fix comment formatting and typos.

Screenshot 2023-10-04 at 2 39 51 PM
Screenshot 2023-10-04 at 3 15 22 PM
Screenshot 2023-10-04 at 3 21 01 PM

Note: `instantswap.ExchangeRateInfo.EstimatedAmount` was previously unused but it gives the correct values for it's current use so I'm using it instead of `instantswap.ExchangeRateInfo.ExchangeRate`.


dreacot:
binance rate still shows on the order scheduler modal

Screenshot from 2023-10-04 11-30-23

libwallet/ext/service.go
```diff
@@ -317,6 +317,12 @@ func (s *Service) GetXpub(xPub string) (xPubBalAndTxs *XpubBalAndTxs, err error)
 func (s *Service) GetTicker(exchange string, market string) (ticker *Ticker, err error) {
 	switch exchange {
 	case Binance:
+		// Return early for dcr-ltc markets ad binance does not support this
```
dreacot, line 320:
```suggestion
		// Return early for dcr-ltc markets as binance does not support this
```

libwallet/instantswap.go
```diff
@@ -103,17 +106,22 @@ func (mgr *AssetsManager) StartScheduler(ctx context.Context, params instantswap
...
-		log.Info(params.Order.ExchangeServer.Server+" rate: ", exchangeServerRate)
-		log.Info(ext.Binance+" rate: ", binanceRate)
+		/// Binance always returns ticker.LastTradePrice in's the quote asset
+		// unit e.g DCR-BTC, LTC-BTC. We will also do this when and if USDT is supported.
+		if strings.EqualFold(fromCur, "btc") {
+			binanceRate = 1 / ticker.LastTradePrice
+		}
+
+		log.Info(params.Order.ExchangeServer.Server+" rate: 1 %s ~= %f %s", fromCur, exchangeServerRate, toCur)
+		log.Info(ext.Binance+" rate: 1 %s ~= %f %s", fromCur, binanceRate, toCur)
```
dreacot, line 124: this can be localized, seems you already have a string entry for it

dreacot (changes requested):

ukane-philemon:
> binance rate still shows on the order scheduler modal

Nice catch. I was unable to sync LTC atm and could not test without syncing.

dreacot (approved):
LGTM!

PR #151, opened 2023-10-05
image: Add layout cache.

### LINKED ISSUE #101: the overall app feels very laggy since the addition of the new layout

JoeGruffins:
Scaling large png backgrounds takes up a lot of resources. Add a cache and only create new images when a different size of the same image is requested.

closes #101

### DIFF
--- a/ui/cryptomaterial/image.go
+++ b/ui/cryptomaterial/image.go
@@ -2,6 +2,7 @@ package cryptomaterial
 
 import (
 	"image"
+	"sync"
 
 	"gioui.org/op/paint"
 	"gioui.org/unit"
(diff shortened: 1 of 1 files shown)


JoeGruffins:
cpu use before and after. This is at the start of the app so I think the benefits from the cache would show more if the app ran longer. Running the app you can definitely feel the difference.

Before:
profile002

After:
profile003

ukane-philemon:
> Running the app you can definitely feel the difference.

Yeah, I do.

dreacot (approved):

dreacot:
Lovely work!

PR #154, opened 2023-10-06
home: Reverse sync toggle logic.

JoeGruffins:
Allows the toggle to start sync.

### DIFF
--- a/ui/page/root/home_page.go
+++ b/ui/page/root/home_page.go
@@ -49,7 +49,7 @@ type HomePage struct {
 
 	// page state variables
 	isBalanceHidden bool
-	isConnected     bool
+	isConnected     *atomic.Bool
 
 	startSpvSync uint32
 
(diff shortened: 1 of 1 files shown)


ui/page/root/home_page.go
```diff
@@ -97,12 +98,12 @@ func NewHomePage(l *load.Load) *HomePage {
...
 
 	go func() {
-		hp.isConnected = libutils.IsOnline()
+		hp.isConnected.Store(libutils.IsOnline())
 	}()
 
 	// init shared page functions
 	toggleSync := func(unlock load.NeedUnlockRestore) {
-		if !hp.WL.SelectedWallet.Wallet.IsConnectedToNetwork() {
+		if hp.WL.SelectedWallet.Wallet.IsConnectedToNetwork() {
```
JoeGruffins, line 106: This is the only change needed... but please have a look at the rest as I think it was missing some thread safety for the isConnected bool and a way out of the connecting loop if ctx is canceled.

dreacot (approved):
LGTM

Issue #140, opened 2023-10-02
Overview Slider image Size Bug


Sirmorrison:
Currently, on the overview coin slider image layout looks like this.
![image](https://github.com/crypto-power/cryptopower/assets/27733432/70ba5037-8164-4f89-a599-2a82aac5289d)

A fix I will like to recommend will be for the coin image to be wrapped by a card or linearLayout that way, regardless of the image size, the image adoptes the size of what ever layout it is in. 

Right now, the image is extending beyond the layout it is in.

JoeGruffins:
What should it look like? A rectangle? Or should the edges be round like on the left? Rounder?

dreacot:
this is what it should look like

![Image](https://github.com/crypto-power/cryptopower/assets/25265396/a1e3e84a-908e-44e0-a196-ba4eb7cc9ea0)

PR #118, opened 2023-09-27
Overview mixer card

crux25:
closes https://github.com/crypto-power/cryptopower/issues/83

This MR adds live data to the mixer card on the overview page.

PR #155, opened 2023-10-06
layout overview page for mobile

### LINKED ISSUE #125: [Mobile]: implement the overview page design
use figma!
Overview page Decred

dreacot:
closes #125 

This MR lays out the overview page for mobile, it does not include api/live data implementations, those would be broken down into different PRs. also the desktop is yet to implement live data for some cards, once those are done there would be follow up issues to populate the cards respective cards. The aim of this PR is simply to provide overview layout for the mobile apps


dreacot:
@Sirmorrison pointed out an issues with the asset balance slider https://github.com/crypto-power/cryptopower/pull/118#issuecomment-1740561191 that causes the radius to look weird

JoeGruffins:
> @Sirmorrison pointed out an issues with the asset balance slider https://github.com/crypto-power/cryptopower/pull/118#issuecomment-1740561191 that causes the radius to look weird

Also cannot flick the slider to see other coins.

dreacot:
> > @Sirmorrison pointed out an issues with the asset balance slider [#118 (comment)](https://github.com/crypto-power/cryptopower/pull/118#issuecomment-1740561191) that causes the radius to look weird
> 
> Also cannot flick the slider to see other coins.

Actually the slider shows the total balance for all asset type, e.g (if you have 3 dcr wallets, it sums up the balance and displays on the dcr asset slider, if you have no other btc/ltc, you won't see the sliders for btc/ltc).

But if you're referring to a fluid transition between sliders, when you flick them, then yes, that would make a lot of sense and would be desirable. We opted not to implement that initially for simplicity.

But I'd create an issue so the sliders can be enhanced

dreacot:
> @Sirmorrison pointed out an issue with the asset balance slider [#118 (comment)](https://github.com/crypto-power/cryptopower/pull/118#issuecomment-1740561191) that causes the radius to look weird

this is the issue in question https://github.com/crypto-power/cryptopower/issues/140

PR #166, opened 2023-10-10
add new layout for wallet selector

dreacot:
closes #145 
closes #162

Screenshot from 2023-10-13 11-00-43
Screenshot from 2023-10-13 11-00-30

PR #175, opened 2023-10-11
add segmented control layout

dreacot:
This PR adds segmented control to the `cryptomaterial` package, i was working on this as part of https://github.com/crypto-power/cryptopower/pull/166, but due to some changes to that PR and the fact that several other layouts would be needing it, i decided to add it as a standalone implementation.

This PR does not include the dragging animation which some segmented controls have, if it is desired it can be added at a later time. but for the sake of avoiding unnecessary delays, i have intentionally left it out

Screenshot from 2023-10-11 03-03-51

### DIFF
--- a/ui/cryptomaterial/segmented_control.go
+++ b/ui/cryptomaterial/segmented_control.go
@@ -1,0 +1,103 @@
+package cryptomaterial
+
+import (
+	"gioui.org/font"
+	"gioui.org/layout"
+
+	"github.com/crypto-power/cryptopower/ui/values"
+)
+
+type SegmentedControl struct {
+	theme *Theme
+	list  *ClickableList
+
+	selectedIndex int
+	segmentItems  []string
+
+	changed bool
+}
+
+func (t *Theme) SegmentedControl(segmentItems []string) *SegmentedControl {
+	list := t.NewClickableList(layout.Horizontal)
+	list.IsHoverable = false
+
+	return &SegmentedControl{
+		list:         list,
+		theme:        t,
+		segmentItems: segmentItems,
+	}
+}
+
+func (sc *SegmentedControl) Layout(gtx C) D {
+	sc.handleEvents()
+
+	return LinearLayout{
+		Width:      WrapContent,
+		Height:     WrapContent,
+		Background: sc.theme.Color.Background,
+		Border:     Border{Radius: Radius(8)},
+	}.Layout(gtx,
+		layout.Rigid(func(gtx C) D {
+			return sc.list.Layout(gtx, len(sc.segmentItems), func(gtx C, i int) D {
+				isSelectedSegment := sc.SelectedIndex() == i
+				return layout.Stack{Alignment: layout.Center}.Layout(gtx,
+					layout.Stacked(func(gtx C) D {
+						return layout.Inset{}.Layout(gtx, func(gtx C) D {
+							return layout.Center.Layout(gtx, func(gtx C) D {
+								bg := sc.theme.Color.SurfaceHighlight
+								txt := sc.theme.Label(values.TextSize16, sc.segmentItems[i])
+								txt.Color = sc.theme.Color.GrayText1
+								txt.Font.Weight = font.SemiBold
+								border := Border{Radius: Radius(0)}
+								if isSelectedSegment {
+									bg = sc.theme.Color.Surface
+									txt.Color = sc.theme.Color.Text
+									border = Border{Radius: Radius(8)}
+								}
+								return LinearLayout{
+									Width:      WrapContent,
+									Height:     WrapContent,
+									Padding:    layout.UniformInset(values.MarginPadding8),
+									Background: bg,
+									Margin:     layout.UniformInset(values.MarginPadding5),
+									Border:     border,
+								}.Layout(gtx,
+									layout.Rigid(func(gtx C) D {
+										return txt.Layout(gtx)
+									}),
+								)
+							})
+						})
+					}),
+				)
+			})
+		}),
+	)
+}
+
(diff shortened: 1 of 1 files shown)


ui/cryptomaterial/segmented_control.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+func (t *Theme) SegmentedControl(segmentItems []string) *SegmentedControl {
+	list := t.NewClickableList(layout.Horizontal)
+	list.IsHoverable = false
+
+	return &SegmentedControl{
+		list:         list,
+		theme:        t,
+		segmentItems: segmentItems,
```
itswisdomagain, line 27: segmentTitles?

ui/cryptomaterial/segmented_control.go
```diff
@@ -0,0 +1,107 @@
...
+		Background: sc.theme.Color.Background,
+		Border:     Border{Radius: Radius(8)},
+	}.Layout(gtx,
+		layout.Rigid(func(gtx C) D {
+			return sc.list.Layout(gtx, len(sc.segmentTitles), func(gtx C, i int) D {
+				isSelectedSegment := sc.SelectedIndex() == i
+				return layout.Stack{Alignment: layout.Center}.Layout(gtx,
+					layout.Stacked(func(gtx C) D {
+						return layout.Inset{}.Layout(gtx, func(gtx C) D {
+							return layout.Center.Layout(gtx, func(gtx C) D {
```
itswisdomagain, line 46: The main widget that displays the button at index i is this last `layout.Center.Layout` line. What's the idea behind placing it inside an empty inset that is stacked inside a stack layout?

Sirmorrison:
Do we really need a segmented control widget? we currently have a widget similar to this that is used on the send page.
`cryptomaterial.SwitchButtonText`.

dreacot:
They are both different control layouts, what you are referring to is a button. The segmented control is used several places in the app, seems reasonable to make it a widget

ui/cryptomaterial/segmented_control.go
```diff
@@ -0,0 +1,107 @@
...
+									Height:     WrapContent,
+									Padding:    layout.UniformInset(values.MarginPadding8),
+									Background: bg,
+									Margin:     layout.UniformInset(values.MarginPadding5),
+									Border:     border,
+								}.Layout(gtx,
+									layout.Rigid(func(gtx C) D {
+										return txt.Layout(gtx)
+									}),
+								)
```
itswisdomagain, line 68: This is basically a text with padding, margin, background and a border. Can we have a custom widget for this? We can probably call it `Label` or `DecoratedText` or something; it can take a background color, padding, margin and border. Even text color and font as well.

itswisdomagain (commented):
Looks okay, but could be improved maybe?

crux25 (commented):
Except for what has been pointed out by other reviewers, the PR LGTM.

ui/cryptomaterial/segmented_control.go
```diff
@@ -0,0 +1,107 @@
...
+}
+
+func (sc *SegmentedControl) SetSelectedSegment(segment string) {
+	for i, item := range sc.segmentTitles {
+		if item == segment {
+			sc.selectedIndex = i
+			break
+		}
+	}
+}
```
JoeGruffins, line 107: I'm new to all this but are `sc.changed` and `sc.selectedIndex` thread safe? Is that a worry for these UI functions or can these functions only happen one after the other? For example if `SelectedSegment()` and `SetSelectedSegment(segment string)` were called at the same time, that would cause a race on `sc.selectedIndex`.

ui/cryptomaterial/segmented_control.go
dreacot: i had the idea of passing an array of widgets as an argument, to the segmented control widget, the array of widget was meant to be the children of the segmented control, hence the stack layout, i didn't go through with that idea anymore due to the complexity.

i've cleaned up that section of code now

ui/cryptomaterial/segmented_control.go
dreacot: good catch, i've added mutexes to the affected functions

ui/cryptomaterial/segmented_control.go
```diff
@@ -0,0 +1,112 @@
...
+						Height:     WrapContent,
+						Padding:    layout.UniformInset(values.MarginPadding8),
+						Background: bg,
+						Margin:     layout.UniformInset(values.MarginPadding5),
+						Border:     border,
+					}.Layout(gtx,
+						layout.Rigid(func(gtx C) D {
+							return txt.Layout(gtx)
+						}),
+					)
```
Sirmorrison, line 66:
```suggestion
					}.Layout2(gtx, txt.Layout)
```

Sirmorrison (changes requested):
This and we are good to go.

Sirmorrison:
Since we are now using this, we will need to depreciate the switch text widget. So I recommend using this widget on the send page so we can test the implementation so far.

dreacot:
> This PR adds segmented control to the cryptomaterial package, i was working on this as part of https://github.com/crypto-power/cryptopower/pull/166, but due to some changes to that PR and the fact that several other layouts would be needing it, i decided to add it as a standalone implementation.

based on my original comment, i brought  the segmented control widget out of my previous PR to make it non blocking. so trying to use it on another page takes us back the original problem

> Since we are now using this, we will need to depreciate the switch text widget. So I recommend using this widget on the send page so we can test the implementation so far.

correct, the text widget would be deprecated, but not as part of this PR. the work on the send page can be done in another PR

Sirmorrison (approved):

PR #177, opened 2023-10-12
multi: Implement RateSource for exchange rates

### LINKED ISSUE #78: implement api to show market overview on the overview page
1. by the top right of the card, there is an option to refresh the data, and see the last refresh rate. refreshing this data manually should also update the USD values for all assets. If the USD value for all assets is being updated for whatever reason, the market overview should also be update, since they are all market related data

2. if the fetch exchange rate api is enabled, then this card becomes visible, if the api is disabled, then this card becomes hidden

the market order should be;
Decred
Bitcoin
Litecoin

Screenshot from 2023-10-03 09-27-50

ukane-philemon:
https://github.com/crypto-power/cryptopower/assets/57448127/5e4e95b6-00bf-4d35-a47e-a5d6cdf16ae4

Closes #78

### DIFF
--- a/libwallet/assets_config.go
+++ b/libwallet/assets_config.go
@@ -6,6 +6,7 @@ import (
 	"decred.org/dcrwallet/v3/errors"
 	"github.com/asdine/storm"
 	"github.com/crypto-power/cryptopower/libwallet/utils"
+	"github.com/crypto-power/cryptopower/ui/values"
 
 	sharedW "github.com/crypto-power/cryptopower/libwallet/assets/wallet"
 
@@ -139,14 +140,25 @@ func (mgr *AssetsManager) GetCurrencyConversionExchange() string {
 	var key string
 	mgr.db.ReadWalletConfigValue(sharedW.CurrencyConversionConfigKey, &key)
 	if key == "" {
-		return "none" // default exchange value
+		return values.DefaultExchangeValue // default exchange value
 	}
 	return key
 }
 
 // SetCurrencyConversionExchange sets the currency conversion exchange.
-func (mgr *AssetsManager) SetCurrencyConversionExchange(data string) {
-	mgr.db.SaveWalletConfigValue(sharedW.CurrencyConversionConfigKey, data)
+func (mgr *AssetsManager) SetCurrencyConversionExchange(xc string) {
+	mgr.db.SaveWalletConfigValue(sharedW.CurrencyConversionConfigKey, xc)
+	go func() {
+		err := mgr.RateSource.ToggleSource(xc)
+		if err != nil {
+			log.Errorf("Failed to toggle rate source: %v", err)
+			return
+		}
+
+		if xc != values.DefaultExchangeValue {
+			mgr.RateSource.Refresh(true)
+		}
+	}()
 }
 
 // GetLanguagePreference returns the language preference.
@@ -188,6 +200,10 @@ func (mgr *AssetsManager) SetTransactionsNotifications(data bool) {
 // SetPrivacyMode sets the privacy mode for the wallet.
 func (mgr *AssetsManager) SetPrivacyMode(isActive bool) {
 	mgr.db.SaveWalletConfigValue(sharedW.PrivacyModeConfigKey, isActive)
+	mgr.RateSource.ToggleStatus(isActive)
+	if !isActive && mgr.GetCurrencyConversionExchange() != values.DefaultExchangeValue {
+		go mgr.RateSource.Refresh(true)
+	}
 }
 
 // IsPrivacyModeOn checks if the privacy mode is set.
--- a/libwallet/assets_config.go
+++ b/libwallet/assets_config.go
@@ -152,11 +152,6 @@ func (mgr *AssetsManager) SetCurrencyConversionExchange(xc string) {
 		err := mgr.RateSource.ToggleSource(xc)
 		if err != nil {
 			log.Errorf("Failed to toggle rate source: %v", err)
-			return
-		}
-
-		if xc != values.DefaultExchangeValue {
-			mgr.RateSource.Refresh(true)
 		}
 	}()
 }
--- a/libwallet/ext/rate_source.go
+++ b/libwallet/ext/rate_source.go
@@ -1,0 +1,892 @@
+// Copyright (c) 2019-2021, The Decred developers
+// Copyright (c) 2023, The Cryptopower developers
+// See LICENSE for details.
+
+package ext
+
+import (
+	"bytes"
+	"compress/flate"
+	"context"
+	"encoding/base64"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"strconv"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/crypto-power/cryptopower/libwallet/utils"
+	"github.com/crypto-power/cryptopower/ui/values"
+)
+
+const (
+	// These are constants used to represent various rate sources supported.
+	bittrex = values.BittrexExchange
+	binance = values.BinanceExchange
+	none    = values.DefaultExchangeValue
+
+	// These are method names for expected bittrex specific websocket messages.
+	BittrexMsgHeartbeat  = "heartbeat"
+	BittrexMarketSummary = "marketSummary"
+	BittrexTicker        = "ticker"
+
+	// MktSep is used repo wide to separate market symbols.
+	MktSep = "-"
+)
+
+var (
+	// These are urls to fetch rate information from the Bittrex exchange.
+	bittrexURLs = sourceURLs{
+		price: "https://api.bittrex.com/v3/markets/%s/ticker",
+		stats: "https://api.bittrex.com/v3/markets/%s/summary",
+		// Bittrex uses SignalR, which retrieves the actual websocket endpoint
+		// via HTTP.
+		ws: "socket-v3.bittrex.com",
+	}
+
+	// These are urls to fetch rate information from the Binance exchange.
+	binanceURLs = sourceURLs{
+		// See: https://binance-docs.github.io/apidocs/spot/en/#current-average-price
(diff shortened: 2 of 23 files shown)


dreacot (changes requested):
```
2023-10-13 16:06:49.807 [INF] DLWL: Loaded 3 wallets
2023-10-13 16:06:49.807 [ERR] CRPW: init assetsManager error: ext.NewCommonRateSource error: New rate source bittrex is not supported
```

my app crashes when i launch your PR

perhaps that error should just be logged and the app allowed to proceed

libwallet/assets_config.go
```diff
@@ -139,14 +140,25 @@ func (mgr *AssetsManager) GetCurrencyConversionExchange() string {
...
 	mgr.db.ReadWalletConfigValue(sharedW.CurrencyConversionConfigKey, &key)
 	if key == "" {
-		return "none" // default exchange value
+		return values.DefaultExchangeValue // default exchange value
 	}
 	return key
 }
 
 // SetCurrencyConversionExchange sets the currency conversion exchange.
-func (mgr *AssetsManager) SetCurrencyConversionExchange(data string) {
```
dreacot, line 148: what do you think of setting the asset rate as a global variable e.g  `AssetRate             map[libutils.AssetType]float64`. and every page that needs to display the exchange rate gets the value from that map. anywhere the rate gets updated, it is update globally too.

ui/page/root/overview_page.go
```diff
@@ -704,17 +807,17 @@ func (pg *OverviewPage) marketTableHeader(gtx C) D {
 			Alignment: layout.Middle,
 		}.Layout(gtx,
 			layout.Flexed(.8, func(gtx C) D {
-				return layout.E.Layout(gtx, pg.assetTableLabel("Price", col))
+				return layout.E.Layout(gtx, pg.assetTableLabel(values.String(values.StrPrice), col))
 			}),
 			layout.Flexed(.2, func(gtx C) D {
-				return layout.E.Layout(gtx, pg.assetTableLabel("ID Change", col))
+				return layout.E.Layout(gtx, pg.assetTableLabel(values.String(values.StrChange), col))
```
dreacot, line 813: This is actually meant to represent 1 day change i.e `1D Change`

dreacot (changes requested):
First pass, solid work!

dreacot:
Below is the Decred icon to be used for that table, it got changed when i implemented the mobile view for the overview page.

the mobile version has a different icon, while the desktop has a different icon, so let's revert this icon without changing the mobile icon

Screenshot from 2023-10-13 16-17-47

ukane-philemon:
> ```
> 2023-10-13 16:06:49.807 [INF] DLWL: Loaded 3 wallets
> 2023-10-13 16:06:49.807 [ERR] CRPW: init assetsManager error: ext.NewCommonRateSource error: New rate source bittrex is not supported
> ```
> 
> 
>     
>   
> 
> my app crashes when i launch your PR
> 
> 
> perhaps that error should just be logged and the app allowed to proceed


Oh, I guess I have to stick to lowercase as we had before now? Unless this change in casing will be an unnecessary pita.

dreacot:
> > ```
> > 2023-10-13 16:06:49.807 [INF] DLWL: Loaded 3 wallets
> > 2023-10-13 16:06:49.807 [ERR] CRPW: init assetsManager error: ext.NewCommonRateSource error: New rate source bittrex is not supported
> > ```
> > 
> > 
> > my app crashes when i launch your PR
> > perhaps that error should just be logged and the app allowed to proceed
> 
> Oh, I guess I have to stick to lowercase as we had before now? Unless this change in casing will be an unnecessary pita.

what change is casing are you referring to?

libwallet/assets_config.go
ukane-philemon: Sounds good. I was thinking about an easy to follow framework for ws updates and reading rates info. A global variable will still require mtx locks in multiple places and ws updates will still be needed.

ukane-philemon:
> what change is casing are you referring to?

Changed the exchange constants value to Title case.

dreacot:
> > what change is casing are you referring to?
> 
> Changed the exchange constants value to Title case.

oh, yeah let's change back to lower case then

ui/page/root/overview_page.go
ukane-philemon: Should we use 24H Change instead? 1D might not be clear for most users, I think.

ukane-philemon:
Desktop look, pending your feedback on 24H(h) Change vs 1D Change @dreacot.

Screenshot 2023-10-14 at 5 36 03 PM

ui/page/root/overview_page.go
dreacot: Yes, this 24H is better

libwallet/assets_config.go
dreacot: Alright then, let's leave as is

dreacot:
> Desktop look, pending your feedback on 24H(h) Change vs 1D Change @dreacot.
> 
> Screenshot 2023-10-14 at 5 36 03 PM

Looks good

ukane-philemon:
https://github.com/crypto-power/cryptopower/pull/177/commits/11af9e481ea5968b2e8ca3757bc11738a8344e56 UI change
Screenshot 2023-10-14 at 6 44 48 PM

dreacot (changes requested):
Looks good overall

Screenshot from 2023-10-17 14-27-05

format the price, so that commas are included, i think there's an `utils.formatUSDBalance` function


also the updated now text keeps updating like every 2 seconds, is that intended?

ukane-philemon:
> also the updated now text keeps updating like every 2 seconds, is that intended? 

we have web sockets implemented so all listeners are notified after each ws message. Still, we can improve by notifying after a certain interval(no longer based on when each message is received).

ukane-philemon:
Updated. 

Screenshot 2023-10-17 at 11 23 05 PM

ui/page/send/page.go
```diff
@@ -408,23 +405,23 @@ func (pg *Page) constructTx() {
...
+func (pg *Page) showBalanceAfterSend() {
 	if pg.sourceAccountSelector != nil {
 		sourceAccount := pg.sourceAccountSelector.SelectedAccount()
 		if sourceAccount.Balance == nil {
 			return
 		}
 		balanceAfterSend := sourceAccount.Balance.Spendable
 		pg.balanceAfterSend = balanceAfterSend.String()
-		pg.balanceAfterSendUSD = utils.FormatUSDBalance(pg.Printer, utils.CryptoToUSD(pg.exchangeRate, balanceAfterSend.ToCoin()))
+		pg.balanceAfterSendUSD = utils.FormatAsUSD(pg.Printer, utils.CryptoToUSD(pg.exchangeRate, balanceAfterSend.ToCoin()))
```
dreacot, line 424: Why was the function name changed, the function took a USD balance and formatted it

ukane-philemon: It was previously limited(in the naming) which is a bit confusing for a generic helper function that's not always used to format a "balance".

dreacot:
> > also the updated now text keeps updating like every 2 seconds, is that intended?
> 
> we have websockets implemented so all listeners are notified after each ws message. Still, we can improve by notifying after a certain interval(no longer based on when each message is received).
> 
> EDIT: Increased the notification interval to 5min for ws updates.

Alright sounds, good

ui/page/send/page.go
dreacot: Are there instances where it was used tl format things other than a balance? Please link me

ukane-philemon: > Are there instances where it was used tl format things other than a balance? Please link me

Not really, but it's now used in the market section in this PR.

dreacot: `FormatAsUSD` doesn't look clear to me.

If the naming concern is due to the function being used to format the price on the market overview, then i think it can fly

ukane-philemon: > FormatAsUSD doesn't look clear to me.

I think I should add "String", yh?

dreacot (approved):

libwallet/ext/rate_source.go
```diff
@@ -0,0 +1,923 @@
...
+
+	// Rates exceeding rateExpiry are expired and should be removed unless there
+	// was an error fetching a new rate.
+	rateExpiry = 30 * time.Minute
+
+	// Rate sources should be refreshed every RateRefreshDuration to replace
+	// expired rates and reconnect websocket if need be.
+	RateRefreshDuration = 60 * time.Minute
+
+	rateNotificationInterval = 5 * time.Minute
```
dreacot, line 80: @ukane-philemon can you confirm this works?, i don't think it does, updated now still happens every 2 seconds

dreacot:
fix conficts

dreacot (approved):

libwallet/instantswap.go
```diff
@@ -311,15 +313,3 @@ func (mgr *AssetsManager) IsOrderSchedulerRunning() bool {
...
-// the from and to currencies used when fetching ticker data from binance.
-func MarketName(fromCur, toCur string) string {
-	dcrToLtc := strings.EqualFold(fromCur, "dcr") && strings.EqualFold(toCur, "ltc")
-	btcToOthersExceptUsdt := strings.EqualFold(fromCur, "btc") && !strings.EqualFold(toCur, "usdt")
-	if dcrToLtc || btcToOthersExceptUsdt {
-		return toCur + "-" + fromCur // e.g DCR/BTC, LTC/BTC and not BTC/LTC or BTC/DCR
-	}
-
-	return fromCur + "-" + toCur
-}
```
dreacot, line 325: @ukane-philemon i see we had a function that did this already, why was this deprecated?

ukane-philemon: Yes. We now have an improved version of this function in the libwallet/ext/rate_source.go file (isSupportedMarket).

JustinBeBoy (approved):
LGTM

PR #179, opened 2023-10-13
add cryptopower to license

### LINKED ISSUE #97: Update Code License
I've noticed that some parts of the code still made reference to Raedah Group when giving credits. I think we should:

1. Update the License file to mention Cryptopower developers (The Cyptopower Developers?)
2. All files should have any applicable License mentioned at the top
3. Copyright dates should be updated.

Vanvians:
Closes #97 
Adds Cryptopower developers to license

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,7 @@
 ISC License
 
 Copyright (c) 2018-2022, Raedah Group
+Copyright (c) 2023, Cryptopower Developers
 
 Permission to use, copy, modify, and/or distribute this software for any
 purpose with or without fee is hereby granted, provided that the above
--- a/LICENSE
+++ b/LICENSE
@@ -1,7 +1,7 @@
 ISC License
 
 Copyright (c) 2018-2022, Raedah Group
-Copyright (c) 2023, Cryptopower Developers
+Copyright (c) 2023, The Cryptopower Developers
 
 Permission to use, copy, modify, and/or distribute this software for any
 purpose with or without fee is hereby granted, provided that the above
--- a/LICENSE
+++ b/LICENSE
@@ -1,7 +1,7 @@
 ISC License
 
 Copyright (c) 2018-2022, Raedah Group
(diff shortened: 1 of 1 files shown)


LICENSE
```diff
@@ -1,6 +1,7 @@
 ISC License
 
 Copyright (c) 2018-2022, Raedah Group
+Copyright (c) 2023, Cryptopower Developers
```
ukane-philemon, line 4: Copyright (c) 2023, The Cryptopower developers

Also, we should start adding the license to each code file starting with this PR. We can do this in batches to ease review. @dreacot what do you think about adding the license to All files that it applies to?

ukane-philemon (changes requested):

LICENSE
dreacot: Copyright (c) 2023, The Cryptopower developers, this can come before the Raedah Group license since it's more current

dreacot: @ukane-philemon your suggestion is fine, can you provide a template you have in mind?

ukane-philemon: > @ukane-philemon your suggestion is fine, can you provide a template you have in mind?

something simple like we have on dcrdata:
> // Copyright (c) 2023 The Cryptopower developers
   // See LICENSE for details.

LICENSE
```diff
@@ -1,7 +1,7 @@
 ISC License
 
 Copyright (c) 2018-2022, Raedah Group
-Copyright (c) 2023, Cryptopower Developers
+Copyright (c) 2023, The Cryptopower Developers
```
ukane-philemon, line 4: all lowercase for developers.

LICENSE
dreacot: sounds good, let's do that with files that have changes to them

ukane-philemon (approved):
Okay, but just a part of #97.

The second part of my https://github.com/crypto-power/cryptopower/pull/179#discussion_r1357805849 was not implemented in this PR but can come in a follow-up PR.

PR #196, opened 2023-10-19
multi: Fix bugs and enable wallet creation on the create order page

### LINKED ISSUE #197: Fix bug on create order page
Noticed a panic on the create order page when a swap asset with zero wallet is selected and a for loop bug that prevents selecting the next swap destination asset.

ukane-philemon:
This fixes a bug that caused a panic on the create order page and makes it a no-op when an asset with zero wallets is selected as a swap asset.

I'm also wondering if we should prompt the user to create the wallet(1), remove the wallet from the list(2) or just do a no-op(MR).

EDIT: I implemented option (1). 

This required moving some reusable files to the components package. I think it's a step a good direction since DEX is coming and we will also need to be able to create missing wallets from the market/main dex page.

Made a vid of the changes and will upload it soon.

closes #197

### DIFF
--- a/ui/page/components/wallet_setup_page.go
+++ b/ui/page/components/wallet_setup_page.go
@@ -635,13 +635,3 @@ func (pg *CreateWallet) validRestoreWalletInputs() bool {
 
 	return true
 }
-
-// func (pg *CreateWallet) walletCreationSuccessCallback() {
-// 	// display the overview page if the user is creating a wallet
-// 	// for the first time (i.e coming from the onboarding page)
-// 	if len(pg.WL.AssetsManager.AllWallets()) == 1 {
-// 		pg.ParentNavigator().Display(NewHomePage(pg.Load))
-// 		return
-// 	}
-// 	pg.ParentNavigator().Display(NewWalletSelectorPage(pg.Load))
-// }
--- a/ui/page/components/wallet_setup_page.go
+++ b/ui/page/components/wallet_setup_page.go
@@ -1,4 +1,4 @@
-package root
+package components
 
 import (
 	"errors"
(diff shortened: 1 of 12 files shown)


dreacot:
it would be best to prompt them to create a wallet

ukane-philemon:
> if so should clicking on the create wallet button open a simple modal (wallet name and spending passphrase), or should it take them to the create wallet page?

I think a modal that has everything they need is an option. Might need a little refactoring so we can use existing code. Clicking on the asset with zero wallet should display the prompt, I think.

dreacot:
> > if so should clicking on the create wallet button open a simple modal (wallet name and spending passphrase), or should it take them to the create wallet page?
> 
> I think a modal that has everything they need is an option. Might need a little refactoring so we can use existing code. Clicking on the asset with zero wallet should display the prompt, I think.

alright then that makes sense

dreacot:
resolve conflicts

ui/page/components/wallet_setup_page.go
```diff
@@ -629,12 +636,12 @@ func (pg *CreateWallet) validRestoreWalletInputs() bool {
...
-}
+// func (pg *CreateWallet) walletCreationSuccessCallback() {
+// 	// display the overview page if the user is creating a wallet
+// 	// for the first time (i.e coming from the onboarding page)
+// 	if len(pg.WL.AssetsManager.AllWallets()) == 1 {
+// 		pg.ParentNavigator().Display(NewHomePage(pg.Load))
+// 		return
+// 	}
+// 	pg.ParentNavigator().Display(NewWalletSelectorPage(pg.Load))
+// }
```
dreacot, line 647: @ukane-philemon remove commented code

PR #198, opened 2023-10-21
Home page transaction entry 

### LINKED ISSUE #180: Create a transactions tab entry entry after the overview tab
create a transactions entry after the overview tab

the UI is meant to replicate what we already have on the wallet transactions page

the transactions tab is meant to be a centralized place to view all transactions combined between assets

there should be a dropdown filter to filter transactions by wallets

image

Sirmorrison:
fix #180 

image

image

image

image

image

### DIFF
--- a/ui/cryptomaterial/dropdown.go
+++ b/ui/cryptomaterial/dropdown.go
@@ -90,7 +90,7 @@ func (d *DropDown) Selected() string {
 	return d.items[d.SelectedIndex()].Text
 }
 
-func (d *DropDown) SetDefaultToNone() {
+func (d *DropDown) ClearSelection() {
 	d.selectedIndex = -1
 }
 
--- a/ui/cryptomaterial/dropdown.go
+++ b/ui/cryptomaterial/dropdown.go
@@ -184,7 +184,7 @@ func (d *DropDown) layoutOption(gtx layout.Context, itemIndex int) D {
 			}
 
 			dropdownItemWidth -= gtx.Dp(values.MarginPadding24) // account for the dropdown Icon
-			return item.Icon.Layout24dp(gtx)
+			return item.Icon.Layout20dp(gtx)
 		}),
 		layout.Rigid(func(gtx C) D {
 			gtx.Constraints.Max.X = dropdownItemWidth - gtx.Dp(values.MarginPadding50) // give some space for the dropdown Icon
--- a/ui/cryptomaterial/dropdown.go
+++ b/ui/cryptomaterial/dropdown.go
@@ -36,6 +36,8 @@ type DropDown struct {
 	linearLayout        *LinearLayout
 	padding             layout.Inset
 	shadow              *Shadow
+
+	noSelectedItemText string
 }
 
 type DropDownItem struct {
@@ -90,8 +92,9 @@ func (d *DropDown) Selected() string {
 	return d.items[d.SelectedIndex()].Text
 }
 
-func (d *DropDown) ClearSelection() {
+func (d *DropDown) ClearSelection(text string) {
 	d.selectedIndex = -1
+	d.noSelectedItemText = text
 }
 
 func (d *DropDown) SelectedIndex() int {
@@ -156,22 +159,22 @@ func (d *DropDown) layoutActiveIcon(gtx layout.Context, index int) D {
 }
 
 func (d *DropDown) layoutOption(gtx layout.Context, itemIndex int) D {
-	var item DropDownItem
+	var item *DropDownItem
 	if itemIndex > -1 {
-		item = d.items[itemIndex]
+		item = &d.items[itemIndex]
 	}
 
 	radius := Radius(8)
 	clickable := d.clickable
 	if d.isOpen {
 		radius = Radius(0)
-		if item.clickable != nil {
+		if item != nil {
 			clickable = item.clickable
 		}
 	}
 
 	padding := values.MarginPadding10
-	if item.Icon != nil {
+	if item != nil && item.Icon != nil {
 		padding = values.MarginPadding8
 	}
 
@@ -189,7 +192,7 @@ func (d *DropDown) layoutOption(gtx layout.Context, itemIndex int) D {
 		Border:    Border{Radius: radius},
 	}.Layout(gtx,
 		layout.Rigid(func(gtx C) D {
-			if item.Icon == nil {
+			if item == nil || item.Icon == nil {
 				return layout.Dimensions{}
 			}
 
@@ -203,11 +206,13 @@ func (d *DropDown) layoutOption(gtx layout.Context, itemIndex int) D {
 				Right: unit.Dp(5),
 				Left:  unit.Dp(5),
 			}.Layout(gtx, func(gtx C) D {
-				txt := item.Text
-				if !d.isOpen && len(txt) > 14 {
+				var txt string
+				if item == nil {
+					txt = d.noSelectedItemText
+				} else if !d.isOpen && len(txt) > 14 {
 					txt = item.Text[:14] + "..."
-				} else if txt == "" {
-					txt = "Select a wallet"
+				} else {
+					txt = item.Text
 				}
 
 				return d.theme.Body2(txt).Layout(gtx)
--- a/ui/cryptomaterial/dropdown.go
+++ b/ui/cryptomaterial/dropdown.go
@@ -90,6 +90,10 @@ func (d *DropDown) Selected() string {
 	return d.items[d.SelectedIndex()].Text
 }
 
+func (d *DropDown) SetDefaultToNone() {
+	d.selectedIndex = -1
+}
+
 func (d *DropDown) SelectedIndex() int {
 	return d.selectedIndex
 }
@@ -152,12 +156,18 @@ func (d *DropDown) layoutActiveIcon(gtx layout.Context, index int) D {
 }
 
 func (d *DropDown) layoutOption(gtx layout.Context, itemIndex int) D {
-	item := d.items[itemIndex]
-	radius := Radius(0)
-	clickable := item.clickable
-	if !d.isOpen {
-		radius = Radius(8)
-		clickable = d.clickable
+	var item DropDownItem
+	if itemIndex > -1 {
+		item = d.items[itemIndex]
+	}
+
+	radius := Radius(8)
+	clickable := d.clickable
+	if d.isOpen {
+		radius = Radius(0)
+		if item.clickable != nil {
+			clickable = item.clickable
+		}
 	}
 
 	padding := values.MarginPadding10
@@ -193,11 +203,14 @@ func (d *DropDown) layoutOption(gtx layout.Context, itemIndex int) D {
 				Right: unit.Dp(5),
 				Left:  unit.Dp(5),
 			}.Layout(gtx, func(gtx C) D {
-				lbl := d.theme.Body2(item.Text)
-				if !d.isOpen && len(item.Text) > 14 {
-					lbl.Text = item.Text[:14] + "..."
+				txt := item.Text
+				if !d.isOpen && len(txt) > 14 {
+					txt = item.Text[:14] + "..."
+				} else if txt == "" {
+					txt = "Select a wallet"
 				}
-				return lbl.Layout(gtx)
+
+				return d.theme.Body2(txt).Layout(gtx)
 			})
 		}),
 		layout.Rigid(func(gtx C) D {
--- a/ui/cryptomaterial/segmented_control.go
+++ b/ui/cryptomaterial/segmented_control.go
@@ -42,7 +42,7 @@ func (sc *SegmentedControl) Layout(gtx C) D {
 	return LinearLayout{
 		Width:      WrapContent,
 		Height:     WrapContent,
-		Background: sc.theme.Color.Background,
+		Background: sc.theme.Color.Gray2,
 		Border:     Border{Radius: Radius(8)},
 	}.Layout(gtx,
 		layout.Rigid(func(gtx C) D {
--- a/ui/page/components/asset_selector.go
+++ b/ui/page/components/asset_selector.go
@@ -179,7 +179,7 @@ func (ats *AssetTypeSelector) Layout(window app.WindowNavigator, gtx C) D {
 			}.Layout(gtx, ats.selectedAssetType.Icon.Layout24dp)
 		}),
 		layout.Rigid(func(gtx C) D {
-			txt := ats.Theme.Label(values.TextSize16, ats.hint)
+			txt := ats.Theme.Label(values.TextSize14, ats.hint)
 			txt.Color = ats.Theme.Color.Gray7
 			if ats.selectedAssetType != nil {
 				txt = ats.Theme.Label(values.TextSize16, ats.selectedAssetType.Type.String())
--- a/ui/page/components/components.go
+++ b/ui/page/components/components.go
@@ -230,7 +230,11 @@ func TransactionTitleIcon(l *load.Load, wal sharedW.Asset, tx *sharedW.Transacti
 // page. It lays out a transaction's direction, balance, status. isTxPage
 // determines if the transaction should be drawn using the transactions page
 // layout.
+<<<<<<< HEAD
 func LayoutTransactionRow(gtx layout.Context, l *load.Load, wal sharedW.Asset, tx *sharedW.Transaction, isTxPage bool) layout.Dimensions {
+=======
+func LayoutTransactionRow(gtx layout.Context, l *load.Load, row TransactionRow, isTxPage bool) layout.Dimensions {
+>>>>>>> 46daad8f (remove asset name)
 	gtx.Constraints.Min.X = gtx.Constraints.Max.X
 	if wal == nil {
 		return D{}
--- a/ui/page/components/components.go
+++ b/ui/page/components/components.go
@@ -352,10 +352,6 @@ func LayoutTransactionRow(gtx layout.Context, l *load.Load, wal sharedW.Asset, t
 								)
 							}
 
-							if isTxPage {
-								return D{}
-							}
-
 							walBalTxt := l.Theme.Label(values.TextSize12, amount)
 							walBalTxt.Color = l.Theme.Color.GrayText2
 							return walBalTxt.Layout(gtx)
--- a/ui/page/components/components.go
+++ b/ui/page/components/components.go
@@ -661,7 +661,7 @@ func TxPageDropDownFields(wType libutils.AssetType, tabIndex int) (mapInfo map[s
 			values.String(values.StrSent),
 			values.String(values.StrReceived),
 		}
-	case wType == libutils.DCRWalletAsset, wType == libutils.NilAsset && tabIndex == 0:
+	case tabIndex == 0 && wType == libutils.DCRWalletAsset, wType == libutils.NilAsset:
 		// DCR Transactions Activities dropdown fields.
 		mapInfo = map[string]int32{
 			values.String(values.StrAll):         libutils.TxFilterAllTx,
--- a/ui/page/components/components.go
+++ b/ui/page/components/components.go
@@ -230,11 +230,7 @@ func TransactionTitleIcon(l *load.Load, wal sharedW.Asset, tx *sharedW.Transacti
 // page. It lays out a transaction's direction, balance, status. isTxPage
 // determines if the transaction should be drawn using the transactions page
 // layout.
-<<<<<<< HEAD
 func LayoutTransactionRow(gtx layout.Context, l *load.Load, wal sharedW.Asset, tx *sharedW.Transaction, isTxPage bool) layout.Dimensions {
-=======
-func LayoutTransactionRow(gtx layout.Context, l *load.Load, row TransactionRow, isTxPage bool) layout.Dimensions {
->>>>>>> 46daad8f (remove asset name)
 	gtx.Constraints.Min.X = gtx.Constraints.Max.X
 	if wal == nil {
 		return D{}
--- a/ui/page/components/components.go
+++ b/ui/page/components/components.go
@@ -227,10 +227,11 @@ func TransactionTitleIcon(l *load.Load, wal sharedW.Asset, tx *sharedW.Transacti
 }
 
 // LayoutTransactionRow is a single transaction row on the transactions and overview
-// page. It lays out a transaction's direction, balance, status. isTxPage
-// determines if the transaction should be drawn using the transactions page
-// layout.
-func LayoutTransactionRow(gtx layout.Context, l *load.Load, wal sharedW.Asset, tx *sharedW.Transaction, isTxPage bool) layout.Dimensions {
+// page. It lays out a transaction's direction, balance, status. isHiddenTxAssetInfo
+// determines if the transaction should display additional information about the tx
+// such as the wallet the tx belong to etc. This is usefil on pages where
+// the tx is displayed from multi wallets.
+func LayoutTransactionRow(gtx layout.Context, l *load.Load, wal sharedW.Asset, tx *sharedW.Transaction, isHiddenTxAssetInfo bool) layout.Dimensions {
 	gtx.Constraints.Min.X = gtx.Constraints.Max.X
 	if wal == nil {
 		return D{}
@@ -242,7 +243,7 @@ func LayoutTransactionRow(gtx layout.Context, l *load.Load, wal sharedW.Asset, t
 	walName := l.Theme.Label(values.TextSize12, wal.GetWalletName())
 
 	insetLeft := values.MarginPadding16
-	if !isTxPage {
+	if !isHiddenTxAssetInfo {
 		insetLeft = values.MarginPadding8
 	}
 
@@ -283,7 +284,7 @@ func LayoutTransactionRow(gtx layout.Context, l *load.Load, wal sharedW.Asset, t
 					}.Layout(gtx,
 						layout.Rigid(l.Theme.Label(values.TextSize18, txStatus.Title).Layout),
 						layout.Rigid(func(gtx C) D {
-							if isTxPage {
+							if isHiddenTxAssetInfo {
 								return D{}
 							}
 							return layout.E.Layout(gtx, func(gtx C) D {
@@ -300,7 +301,7 @@ func LayoutTransactionRow(gtx layout.Context, l *load.Load, wal sharedW.Asset, t
 					)
 				}),
 				layout.Rigid(func(gtx C) D {
-					if !isTxPage && tx.Type == txhelper.TxTypeRegular {
+					if !isHiddenTxAssetInfo && tx.Type == txhelper.TxTypeRegular {
 						return cryptomaterial.LinearLayout{
 							Width:       cryptomaterial.WrapContent,
 							Height:      cryptomaterial.WrapContent,
@@ -322,7 +323,7 @@ func LayoutTransactionRow(gtx layout.Context, l *load.Load, wal sharedW.Asset, t
 						Direction:   layout.W,
 					}.Layout(gtx,
 						layout.Rigid(func(gtx C) D {
-							if isTxPage {
+							if isHiddenTxAssetInfo {
 								return D{}
 							}
 
@@ -357,7 +358,7 @@ func LayoutTransactionRow(gtx layout.Context, l *load.Load, wal sharedW.Asset, t
 							return walBalTxt.Layout(gtx)
 						}),
 						layout.Rigid(func(gtx C) D {
-							if dcrAsset, ok := wal.(*dcr.Asset); ok && !isTxPage {
+							if dcrAsset, ok := wal.(*dcr.Asset); ok && !isHiddenTxAssetInfo {
 								if ok, _ := dcrAsset.TicketHasVotedOrRevoked(tx.Hash); ok {
 									return layout.Inset{
 										Left: values.MarginPadding4,
@@ -376,7 +377,7 @@ func LayoutTransactionRow(gtx layout.Context, l *load.Load, wal sharedW.Asset, t
 								ticketSpender, _ = dcrAsset.TicketSpender(tx.Hash)
 							}
 
-							if ticketSpender == nil || isTxPage {
+							if ticketSpender == nil || isHiddenTxAssetInfo {
 								return D{}
 							}
 							amnt := wal.ToAmount(ticketSpender.VoteReward).ToCoin()
@@ -392,7 +393,7 @@ func LayoutTransactionRow(gtx layout.Context, l *load.Load, wal sharedW.Asset, t
 		}),
 		layout.Flexed(1, func(gtx C) D {
 			txSize := values.TextSize16
-			if !isTxPage {
+			if !isHiddenTxAssetInfo {
 				txSize = values.TextSize12
 			}
 			status := l.Theme.Label(txSize, values.String(values.StrUnknown))
@@ -415,7 +416,7 @@ func LayoutTransactionRow(gtx layout.Context, l *load.Load, wal sharedW.Asset, t
 				return layout.Flex{Alignment: layout.Baseline}.Layout(gtx,
 					layout.Rigid(func(gtx C) D {
 						voteOrRevocationTx := tx.Type == txhelper.TxTypeVote || tx.Type == txhelper.TxTypeRevocation
-						if isTxPage && voteOrRevocationTx {
+						if isHiddenTxAssetInfo && voteOrRevocationTx {
 							title := values.String(values.StrRevoke)
 							if tx.Type == txhelper.TxTypeVote {
 								title = values.String(values.StrVote)
@@ -443,7 +444,7 @@ func LayoutTransactionRow(gtx layout.Context, l *load.Load, wal sharedW.Asset, t
 						return D{}
 					}),
 					layout.Rigid(func(gtx C) D {
-						if !isTxPage {
+						if !isHiddenTxAssetInfo {
 							return cryptomaterial.LinearLayout{
 								Width:       cryptomaterial.WrapContent,
 								Height:      cryptomaterial.WrapContent,
@@ -481,14 +482,14 @@ func LayoutTransactionRow(gtx layout.Context, l *load.Load, wal sharedW.Asset, t
 						return D{}
 					}),
 					layout.Rigid(func(gtx C) D {
-						if isTxPage {
+						if isHiddenTxAssetInfo {
 							return status.Layout(gtx)
 						}
 						return D{}
 					}),
 					layout.Rigid(func(gtx C) D {
 						isMixedOrRegular := tx.Type == txhelper.TxTypeMixed || tx.Type == txhelper.TxTypeRegular
-						if !isTxPage && !isMixedOrRegular {
+						if !isHiddenTxAssetInfo && !isMixedOrRegular {
 							return D{}
 						}
 						statusIcon := l.Theme.Icons.ConfirmIcon
@@ -497,7 +498,7 @@ func LayoutTransactionRow(gtx layout.Context, l *load.Load, wal sharedW.Asset, t
 							statusIcon = l.Theme.Icons.PendingIcon
 						}
 
-						if isTxPage {
+						if isHiddenTxAssetInfo {
 							return layout.Inset{
 								Left: values.MarginPadding15,
 								Top:  values.MarginPadding5,
@@ -627,7 +628,7 @@ func TimeFormat(secs int, long bool) string {
 
 // TxPageDropDownFields returns the fields for the required drop down with the
 // transactions view page. Since maps access of items order is always random
-// an array of keys is provided guarrantee the dropdown order will always be
+// an array of keys is provided guarantee the dropdown order will always be
 // maintained.
 func TxPageDropDownFields(wType libutils.AssetType, tabIndex int) (mapInfo map[string]int32, keysInfo []string) {
 	switch {
@@ -655,7 +656,7 @@ func TxPageDropDownFields(wType libutils.AssetType, tabIndex int) (mapInfo map[s
 			values.String(values.StrSent),
 			values.String(values.StrReceived),
 		}
-	case tabIndex == 0 && wType == libutils.DCRWalletAsset, wType == libutils.NilAsset:
+	case wType == libutils.DCRWalletAsset && tabIndex == 0:
 		// DCR Transactions Activities dropdown fields.
 		mapInfo = map[string]int32{
 			values.String(values.StrAll):         libutils.TxFilterAllTx,
@@ -683,8 +684,6 @@ func TxPageDropDownFields(wType libutils.AssetType, tabIndex int) (mapInfo map[s
 			values.String(values.StrVote),
 			values.String(values.StrRevocation),
 		}
-	default:
-
 	}
 	return
 }
--- a/ui/page/components/components.go
+++ b/ui/page/components/components.go
@@ -226,17 +226,11 @@ func TransactionTitleIcon(l *load.Load, wal sharedW.Asset, tx *sharedW.Transacti
 	return &txStatus
 }
 
-<<<<<<< HEAD
-// transactionRow is a single transaction row on the transactions and overview page. It lays out a transaction's
-// direction, balance, status. isTxPage determines if the transaction should be drawn using the transactions page layout.
-func LayoutTransactionRow(gtx layout.Context, l *load.Load, wal sharedW.Asset, tx *sharedW.Transaction, isTxPage bool) layout.Dimensions {
-=======
 // transactionRow is a single transaction row on the transactions and overview
 // page. It lays out a transaction's direction, balance, status. isTxPage
 // determines if the transaction should be drawn using the transactions page
 // layout.
-func LayoutTransactionRow(gtx layout.Context, l *load.Load, selectedWallet *load.WalletMapping, row TransactionRow, isTxPage bool) layout.Dimensions {
->>>>>>> 16d25c9b (add tx entry page to the home page)
+func LayoutTransactionRow(gtx layout.Context, l *load.Load, wal sharedW.Asset, tx *sharedW.Transaction, isTxPage bool) layout.Dimensions {
 	gtx.Constraints.Min.X = gtx.Constraints.Max.X
 	if wal == nil {
 		return D{}
@@ -320,23 +314,23 @@ func LayoutTransactionRow(gtx layout.Context, l *load.Load, selectedWallet *load
 							}),
 						)
 					}
-
 					return cryptomaterial.LinearLayout{
 						Width:       cryptomaterial.WrapContent,
 						Height:      cryptomaterial.WrapContent,
 						Orientation: layout.Horizontal,
 						Alignment:   layout.Middle,
+						Direction:   layout.W,
 					}.Layout(gtx,
 						layout.Rigid(func(gtx C) D {
 							if isTxPage {
 								return D{}
 							}
+
 							if tx.Type == txhelper.TxTypeMixed {
 								return cryptomaterial.LinearLayout{
 									Width:       cryptomaterial.WrapContent,
 									Height:      cryptomaterial.WrapContent,
 									Orientation: layout.Horizontal,
-									Direction:   layout.W,
 									Alignment:   layout.Middle,
 								}.Layout(gtx,
 									layout.Rigid(func(gtx C) D {
@@ -502,11 +496,8 @@ func LayoutTransactionRow(gtx layout.Context, l *load.Load, selectedWallet *load
 							return D{}
 						}
 						statusIcon := l.Theme.Icons.ConfirmIcon
-<<<<<<< HEAD
+
 						if TxConfirmations(wal, tx) < wal.RequiredConfirmations() {
-=======
-						if TxConfirmations(l, row.Transaction) < selectedWallet.RequiredConfirmations() {
->>>>>>> 16d25c9b (add tx entry page to the home page)
 							statusIcon = l.Theme.Icons.PendingIcon
 						}
 
--- a/ui/page/components/components.go
+++ b/ui/page/components/components.go
@@ -226,7 +226,7 @@ func TransactionTitleIcon(l *load.Load, wal sharedW.Asset, tx *sharedW.Transacti
 	return &txStatus
 }
 
-// transactionRow is a single transaction row on the transactions and overview
+// LayoutTransactionRow is a single transaction row on the transactions and overview
 // page. It lays out a transaction's direction, balance, status. isTxPage
 // determines if the transaction should be drawn using the transactions page
 // layout.
--- a/ui/page/components/components.go
+++ b/ui/page/components/components.go
@@ -226,9 +226,17 @@ func TransactionTitleIcon(l *load.Load, wal sharedW.Asset, tx *sharedW.Transacti
 	return &txStatus
 }
 
+<<<<<<< HEAD
 // transactionRow is a single transaction row on the transactions and overview page. It lays out a transaction's
 // direction, balance, status. isTxPage determines if the transaction should be drawn using the transactions page layout.
 func LayoutTransactionRow(gtx layout.Context, l *load.Load, wal sharedW.Asset, tx *sharedW.Transaction, isTxPage bool) layout.Dimensions {
+=======
+// transactionRow is a single transaction row on the transactions and overview
+// page. It lays out a transaction's direction, balance, status. isTxPage
+// determines if the transaction should be drawn using the transactions page
+// layout.
+func LayoutTransactionRow(gtx layout.Context, l *load.Load, selectedWallet *load.WalletMapping, row TransactionRow, isTxPage bool) layout.Dimensions {
+>>>>>>> 16d25c9b (add tx entry page to the home page)
 	gtx.Constraints.Min.X = gtx.Constraints.Max.X
 	if wal == nil {
 		return D{}
@@ -494,7 +502,11 @@ func LayoutTransactionRow(gtx layout.Context, l *load.Load, wal sharedW.Asset, t
 							return D{}
 						}
 						statusIcon := l.Theme.Icons.ConfirmIcon
+<<<<<<< HEAD
 						if TxConfirmations(wal, tx) < wal.RequiredConfirmations() {
+=======
+						if TxConfirmations(l, row.Transaction) < selectedWallet.RequiredConfirmations() {
+>>>>>>> 16d25c9b (add tx entry page to the home page)
 							statusIcon = l.Theme.Icons.PendingIcon
 						}
 
--- a/ui/page/components/components.go
+++ b/ui/page/components/components.go
@@ -21,6 +21,7 @@ import (
 
 	"github.com/ararog/timeago"
 	"github.com/crypto-power/cryptopower/app"
+	"github.com/crypto-power/cryptopower/libwallet"
 	"github.com/crypto-power/cryptopower/libwallet/assets/dcr"
 	sharedW "github.com/crypto-power/cryptopower/libwallet/assets/wallet"
 	"github.com/crypto-power/cryptopower/libwallet/txhelper"
@@ -65,6 +66,11 @@ type (
 
 	DexServer struct {
 		SavedHosts map[string][]byte
+	}
+
+	MultiWalletTx struct {
+		*sharedW.Transaction
+		WalletID int
 	}
 )
 
@@ -655,7 +661,7 @@ func TxPageDropDownFields(wType libutils.AssetType, tabIndex int) (mapInfo map[s
 			values.String(values.StrSent),
 			values.String(values.StrReceived),
 		}
-	case wType == libutils.DCRWalletAsset && tabIndex == 0:
+	case wType == libutils.DCRWalletAsset, wType == libutils.NilAsset && tabIndex == 0:
 		// DCR Transactions Activities dropdown fields.
 		mapInfo = map[string]int32{
 			values.String(values.StrAll):         libutils.TxFilterAllTx,
@@ -683,6 +689,8 @@ func TxPageDropDownFields(wType libutils.AssetType, tabIndex int) (mapInfo map[s
 			values.String(values.StrVote),
 			values.String(values.StrRevocation),
 		}
+	default:
+
 	}
 	return
 }
@@ -1001,4 +1009,8 @@ func InputsNotEmpty(editors ...*widget.Editor) bool {
 		}
 	}
 	return true
+}
+
+func TxAndWallet(assetManager *libwallet.AssetsManager, mtx *MultiWalletTx) (*sharedW.Transaction, sharedW.Asset) {
+	return mtx.Transaction, assetManager.WalletWithID(mtx.WalletID)
 }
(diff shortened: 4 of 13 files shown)


dreacot:
Screenshot from 2023-10-23 11-20-53

this should be a wallet dropdown selector

dreacot:
clicking on a tx details causes a crash 

```
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x12e424c]

goroutine 181 [running]:
github.com/crypto-power/cryptopower/ui/page/transaction.NewTransactionDetailsPage(0xc0043f7c20, 0xc005180000, 0x18?)
        /home/kennedy/Projects/src/github.com/cryptopower/cryptopower/ui/page/transaction/transaction_details_page.go:121 +0xd4c
github.com/crypto-power/cryptopower/ui/page/transaction.(*TransactionsPage).HandleUserInteractions(0xc0005b2370)
        /home/kennedy/Projects/src/github.com/cryptopower/cryptopower/ui/page/transaction/transactions_page.go:413 +0xd6
github.com/crypto-power/cryptopower/ui/page/root.(*HomePage).HandleUserInteractions(0xc000004f00)
        /home/kennedy/Projects/src/github.com/cryptopower/cryptopower/ui/page/root/home_page.go:186 +0x67
github.com/crypto-power/cryptopower/ui.(*Window).handleFrameEvent(0xc000189860, {{0xc145c8d933aa1f39, 0x2ad18a474f, 0x2ebb220}, {0x3f800000, 0x3f800000}, {0x320, 0x28a}, {0x0, 0x0, ...}, ...})
        /home/kennedy/Projects/src/github.com/cryptopower/cryptopower/ui/window.go:233 +0x14c
github.com/crypto-power/cryptopower/ui.(*Window).HandleEvents(0xc000189860)
        /home/kennedy/Projects/src/github.com/cryptopower/cryptopower/ui/window.go:206 +0x2b9
main.main.func3()
        /home/kennedy/Projects/src/github.com/cryptopower/cryptopower/main.go:110 +0x1d
created by main.main
        /home/kennedy/Projects/src/github.com/cryptopower/cryptopower/main.go:108 +0x61b
exit status 2
```

JustinBeBoy:
app stuck when change filter to sent.

image

Transferred notice "(0)" but we have data on list.

image

the same above

image

"All" filter notice 2 but empty in list

image

JustinBeBoy:
The right margin needs alignment

image

JustinBeBoy:
App stuck when switch between wallets

image

Please warn users when choosing a wallet that is not synchronized like this:
image

dreacot:
staking activity doesn't show when selected

Screenshot from 2023-10-24 11-40-59

Sirmorrison:
> app stuck when change filter to sent.
> 
> image
> 
> Transferred notice "(0)" but we have data on list.
> 
> image
> 
> the same above
> 
> image
> 
> "All" filter notice 2 but empty in list
> 
> image

I could not reproduce the part where the app gets stuck besides that, I attended to all other comments.

JustinBeBoy:
Layout break when choose  not synced wallet

image

JustinBeBoy:
> The right margin needs alignment
> 
> image

it not resolve

JustinBeBoy (changes requested):
please check my comments

Sirmorrison:
> Layout break when choose not synced wallet
> 
> image

by layout break, how do you mean?

JustinBeBoy:
> > Layout break when choose not synced wallet
> > image
> 
> By layout break, how do you mean? If you mean you just see the red text, thats the design because we cannot using the overlay would involve changing how the overlay is displayed from a selected wallet view.

Sorry, I mean tab's title not center when choose not sync wallet
image

Sirmorrison:
> > The right margin needs alignment
> > image
> 
> it not resolve

This is as a result of the scroll list.. there is an extra padding for the scroll bar. I will see if i can adjust it as its currently on master not my PR.

JustinBeBoy:
> > > The right margin needs alignment
> > > image
> > 
> > 
> > it not resolve
> 
> This is as a result of the scroll list.. there is an extra padding for the scroll bar. I will see if i can adjust it as its currently on master not my PR.

I think we should do it on this PR, so it will be perfect

dreacot:
why not make use of the existing overlay for this situation

Screenshot from 2023-10-26 12-15-40

Screenshot from 2023-10-26 12-17-06

dreacot:
i was on the revocation dropdown on the staking segmented control, switching the segmented control to transaction overview, showed the error seen in the screenshot below

Screenshot from 2023-10-26 12-17-47

dreacot:
trying to switch the wallet selector freezes the app

Screenshot from 2023-10-26 12-25-51

Sirmorrison:
> trying to switch the wallet selector freezes the app
> 
> Screenshot from 2023-10-26 12-25-51

I am unable to reproduce this.

Sirmorrison:
> i was on the revocation dropdown on the staking segmented control, switching the segmented control to transaction overview, showed the error seen in the screenshot below
> 
> Screenshot from 2023-10-26 12-17-47
> 
> and then after clicking okay, the loader just keep showing as seen below
> 
> Screenshot from 2023-10-26 12-19-00

resolved



> why not make use of the existing overlay for this situation
> 
> Screenshot from 2023-10-26 12-15-40
> 
> Screenshot from 2023-10-26 12-17-06

Resolved



> > > > The right margin needs alignment
> > > > image
> > > 
> > > 
> > > it not resolve
> > 
> > 
> > This is as a result of the scroll list.. there is an extra padding for the scroll bar. I will see if i can adjust it as its currently on master not my PR.
> 
> I think we should do it on this PR, so it will be perfect

resolved

dreacot:
the transaction item isn't showing the correct wallet name

Screenshot 2023-10-29 at 2 28 43 PM

dreacot:
> > trying to switch the wallet selector freezes the app
> > Screenshot from 2023-10-26 12-25-51
> 
> I am unable to reproduce this.

i'd get a screen recording, this happens when i select wallets of different assets from the dropdown about 4 times in a row

dreacot:
> > > trying to switch the wallet selector freezes the app
> > > Screenshot from 2023-10-26 12-25-51
> > 
> > 
> > I am unable to reproduce this.
> 
> i'd get a screen recording, this happens when i select wallets of different assets from the dropdown about 4 times in a row


https://github.com/crypto-power/cryptopower/assets/25265396/b42bd5ab-c06d-4d44-8e25-ea05f6d591d8

dreacot:
> > > > trying to switch the wallet selector freezes the app
> > > > Screenshot from 2023-10-26 12-25-51
> > > 
> > > 
> > > I am unable to reproduce this.
> > 
> > 
> > i'd get a screen recording, this happens when i select wallets of different assets from the dropdown about 4 times in a row
> 
>  Screen.Recording.2023-10-29.at.2.44.43.PM.mov

seems like this occurs when btc goes into recovery mode, might not be directly related to this PR

dreacot (approved):

JustinBeBoy:
I still stuck when change tab to Staking Activities
I think it occurs when you have many transactions. I have 5 time. It always occurs

image

ui/cryptomaterial/dropdown.go
```diff
@@ -184,7 +184,7 @@ func (d *DropDown) layoutOption(gtx layout.Context, itemIndex int) D {
 			}
 
 			dropdownItemWidth -= gtx.Dp(values.MarginPadding24) // account for the dropdown Icon
-			return item.Icon.Layout24dp(gtx)
+			return item.Icon.Layout20dp(gtx)
```
itswisdomagain, line 187: I'm weary of changes to reusable components like this. It might make the component look the way you want on the part of the UI you're working on but would affect other parts of the UI using the component.

What's the motivation for this change?

ui/cryptomaterial/segmented_control.go
```diff
@@ -42,7 +42,7 @@ func (sc *SegmentedControl) Layout(gtx C) D {
 	return LinearLayout{
 		Width:      WrapContent,
 		Height:     WrapContent,
-		Background: sc.theme.Color.Background,
+		Background: sc.theme.Color.Gray2,
```
itswisdomagain, line 45: What's the motivation for this change?

ui/page/components/asset_selector.go
```diff
@@ -179,7 +179,7 @@ func (ats *AssetTypeSelector) Layout(window app.WindowNavigator, gtx C) D {
 			}.Layout(gtx, ats.selectedAssetType.Icon.Layout24dp)
 		}),
 		layout.Rigid(func(gtx C) D {
-			txt := ats.Theme.Label(values.TextSize16, ats.hint)
+			txt := ats.Theme.Label(values.TextSize14, ats.hint)
```
itswisdomagain, line 182: What's the motivation for this change?

ui/page/components/components.go
```diff
@@ -230,8 +230,10 @@ func TransactionTitleIcon(l *load.Load, wal sharedW.Asset, tx *sharedW.Transacti
 	return &txStatus
 }
 
-// transactionRow is a single transaction row on the transactions and overview page. It lays out a transaction's
-// direction, balance, status. isTxPage determines if the transaction should be drawn using the transactions page layout.
+// transactionRow is a single transaction row on the transactions and overview
+// page. It lays out a transaction's direction, balance, status. isTxPage
+// determines if the transaction should be drawn using the transactions page
+// layout.
 func LayoutTransactionRow(gtx layout.Context, l *load.Load, row TransactionRow, isTxPage bool) layout.Dimensions {
```
itswisdomagain, line 237: The comment describes "transactionRow" not "LayoutTransactionRow".

ui/page/transaction/transactions_page.go
```diff
@@ -54,27 +51,45 @@ type TransactionsPage struct {
...
+	assetWallets         []sharedW.Asset
+	selectedWallet       sharedW.Asset
+
+	isHomepageLayout,
+	showLoader,
+	showDisabledLayout bool
 }
 
-func NewTransactionsPage(l *load.Load) *TransactionsPage {
+func NewTransactionsPage(l *load.Load, isHomepageLayout bool) *TransactionsPage {
```
itswisdomagain, line 74: This should take a `wallet` instead of taking `isHomepageLayout bool`. If a nil wallet is passed, then it's obvious that the page should show transactions for all wallets. But if a wallet is passed, the page should show transactions for that wallet. We don't care if the caller is the home page or some other page.

ui/page/transaction/transactions_page.go
```diff
@@ -54,27 +51,45 @@ type TransactionsPage struct {
...
 	pg := &TransactionsPage{
 		Load:             l,
 		GenericPageModal: app.NewGenericPageModal(TransactionsPageID),
 		separator:        l.Theme.Separator(),
 		transactionList:  l.Theme.NewClickableList(layout.Vertical),
+		isHomepageLayout: isHomepageLayout,
+	}
+
+	pg.tab = l.Theme.SegmentedControl(txTabs)
```
itswisdomagain, line 83: Assign this value when `pg` is initialized above.

ui/page/transaction/transactions_page.go
```diff
@@ -90,9 +105,8 @@ func NewTransactionsPage(l *load.Load) *TransactionsPage {
 // Part of the load.Page interface.
 func (pg *TransactionsPage) OnNavigatedTo() {
 	pg.ctx, pg.ctxCancel = context.WithCancel(context.TODO())
-
 	pg.refreshAvailableTxType()
-	if !pg.WL.SelectedWallet.Wallet.IsSynced() {
+	if !pg.selectedWallet.IsSynced() {
```
itswisdomagain, line 109: Out of scope for this PR, but what if the wallet becomes synced while this page is open? Can leave a TODO if necessary.

ui/page/transaction/transactions_page.go
```diff
@@ -101,50 +115,42 @@ func (pg *TransactionsPage) OnNavigatedTo() {
...
-					}.Layout(gtx, func(gtx C) D {
-						return layout.Center.Layout(gtx, func(gtx C) D {
-							lbl := pg.Theme.Label(values.TextSize16, txTabs[i])
-							lbl.Color = pg.Theme.Color.GrayText1
-							if i == pg.selectedTabIndex {
-								lbl.Color = pg.Theme.Color.Primary
-								selectedTabDims = lbl.Layout(gtx)
-							}
+// initWalletSelector is used by the tx history page, for wallet selection.
+func (pg *TransactionsPage) initWalletDropdown() {
```
itswisdomagain, line 119: This is the tx history page, what else would use this method!?
```suggestion
// initWalletSelector initializes the wallet selector dropdown to enable filtering
// transactions for a specific wallet when this page is used to display transactions
// for multiple wallets.
func (pg *TransactionsPage) initWalletDropdown() {
```

itswisdomagain: This comment reminds me, I hope we don't show a dropdown if there's only one wallet in the app. If there's only one wallet, it should be auto selected and the page should render as if it were displaying transactions for just one wallet, which it is.

ui/page/transaction/transactions_page.go
itswisdomagain, line 83: This segmented control is only necessary if there is at least 1 dcr wallet in the app. Otherwise, don't show this.

ui/page/transaction/transactions_page.go
```diff
@@ -101,50 +115,42 @@ func (pg *TransactionsPage) OnNavigatedTo() {
...
-						})
-					})
-				}),
-				layout.Stacked(func(gtx C) D {
-					if i != pg.selectedTabIndex {
-						return D{}
-					}
+	items := []cryptomaterial.DropDownItem{}
+	for _, wal := range pg.assetWallets {
+		item := cryptomaterial.DropDownItem{}
```
itswisdomagain, line 128: Introduce a `pg.Theme.AssetIcon()` method to simplify things.

```suggestion
		item := cryptomaterial.DropDownItem{
			Text: wal.GetWalletName(),
			Icon: pg.Theme.AssetIcon(wal.GetAssetType()),
		}
```

ui/page/transaction/transactions_page.go
```diff
@@ -101,50 +115,42 @@ func (pg *TransactionsPage) OnNavigatedTo() {
...
+		}
+		item.Text = wal.GetWalletName()
+		items = append(items, item)
+	}
 
-					tabHeight := gtx.Dp(values.MarginPadding2)
-					tabRect := image.Rect(0, 0, selectedTabDims.Size.X, tabHeight)
+	pg.walletDropDown = pg.Theme.DropDown(items, values.WalletsDropdownGroup, 0)
+	pg.selectedWallet = pg.assetWallets[pg.walletDropDown.SelectedIndex()]
+	pg.showLoader = false
```
itswisdomagain, line 143: What is achieved by setting this to true at the beginning of the method and then changing it to false at the end? It appears this pg.showLoader property is meant to influence the UI layout but UI layout doesn't happen at any point between the start and the end of this method. Changing the pg.showLoader value at the start and end of this method seems moot to me.

ui/page/transaction/transactions_page.go
```diff
@@ -186,11 +193,12 @@ func (pg *TransactionsPage) refreshAvailableTxType() {
 		}
 		pg.txTypeDropDown = pg.Theme.DropDown(items, values.TxDropdownGroup, 2)
 		pg.ParentWindow().Reload()
+		pg.showLoader = false
```
itswisdomagain, line 196: How long does it take for this goroutine to complete? Can we do this in the main thread instead of in the background and then showing a loader until it completes?

ui/page/transaction/transactions_page.go
```diff
@@ -222,95 +230,127 @@ func (pg *TransactionsPage) loadTransactions(offset, pageSize int32) (interface{
...
-func (pg *TransactionsPage) Layout(gtx layout.Context) layout.Dimensions {
+func (pg *TransactionsPage) Layout(gtx C) D {
 	if pg.Load.GetCurrentAppWidth() <= gtx.Dp(values.StartMobileView) {
 		return pg.layoutMobile(gtx)
 	}
 	return pg.layoutDesktop(gtx)
 }
 
-func (pg *TransactionsPage) layoutDesktop(gtx layout.Context) layout.Dimensions {
+func (pg *TransactionsPage) pageLayout(gtx C, body layout.Widget) D {
```
itswisdomagain, line 240: `pg.pageLayout` stutters (pg/page is repeated). Also, did you mean `layoutPage`? There's already a `pg.Layout()` method, so whatever this does, rename the method and add a descriptive comment.

ui/page/transaction/transactions_page.go
```diff
@@ -101,50 +115,42 @@ func (pg *TransactionsPage) OnNavigatedTo() {
...
-					}.Layout(gtx, func(gtx C) D {
-						paint.FillShape(gtx.Ops, pg.Theme.Color.Primary, clip.Rect(tabRect).Op())
-						return layout.Dimensions{
-							Size: image.Point{X: selectedTabDims.Size.X, Y: tabHeight},
-						}
-					})
-				}),
-			)
-		})
+func (pg *TransactionsPage) sectionNavTab(gtx C) D {
```
itswisdomagain, line 146: Move this close to where it is called.

itswisdomagain (changes requested):

ui/cryptomaterial/dropdown.go
Sirmorrison: Icon size to text size. The size of the icon is way too big compared to the text. Makes the UI look like Johnny bravo.

ui/cryptomaterial/segmented_control.go
Sirmorrison: Figma design.

ui/page/components/asset_selector.go
Sirmorrison: Figma design text size.

ui/page/transaction/transactions_page.go
Sirmorrison: Ermmmm... the wallet selection happens inside the Transactions page not the home page. Not sure how to get the wallet that will be passed to `NewtransactionsPage()`

ui/page/transaction/transactions_page.go
Sirmorrison: It auto selects the first wallet in the dropdown, but does not hide the wallet selector is its just one.

ui/cryptomaterial/dropdown.go
Sirmorrison: image
this is what it currently looks like and i still think the icon is still too big but its way better than using 24px.

ui/cryptomaterial/segmented_control.go
Sirmorrison: image
that is gray 2.

ui/page/transaction/transactions_page.go
Sirmorrison: Fixed

ui/cryptomaterial/dropdown.go
itswisdomagain: I see that none of the existing users of the Dropdown were using icon. It's probably why no one noticed that the size was too much.

ui/page/transaction/transactions_page.go
itswisdomagain: When the transactions page is opened from a specific wallet's home page. I think each specific wallet's home page is currently called MainPage.

ui/page/transaction/transactions_page.go
```diff
@@ -222,95 +222,138 @@ func (pg *TransactionsPage) loadTransactions(offset, pageSize int32) (interface{
...
-func (pg *TransactionsPage) layoutMobile(gtx layout.Context) layout.Dimensions {
+func (pg *TransactionsPage) sectionNavTab(gtx C) D {
+	return cryptomaterial.LinearLayout{
+		Width:       cryptomaterial.MatchParent,
+		Height:      cryptomaterial.WrapContent,
+		Orientation: layout.Horizontal,
+		Direction:   layout.Center,
+	}.Layout2(gtx, func(gtx C) D {
+		return layout.Inset{Bottom: values.MarginPadding16}.Layout(gtx, pg.tab.Layout)
+	})
```
itswisdomagain, line 353: I'll like to see a `cryptomaterial.CentralizeWidget(widget D)` method that uses a LinearLayout to do the centralization. We shouldn't be writing plenty code each time just to centralize one widget.

ui/page/transaction/transactions_page.go
```diff
@@ -384,15 +427,29 @@ func (pg *TransactionsPage) HandleUserInteractions() {
...
 	}
-	cryptomaterial.DisplayOneDropdown(pg.txTypeDropDown)
 
-	if tabItemClicked, clickedTabIndex := pg.tabs.ItemClicked(); tabItemClicked {
-		pg.selectedTabIndex = clickedTabIndex
+	dropDownList := []*cryptomaterial.DropDown{pg.txTypeDropDown}
+	if pg.isHomepageLayout {
+		dropDownList = append(dropDownList, pg.walletDropDown)
+	}
+	cryptomaterial.DisplayOneDropdown(dropDownList...)
```
itswisdomagain, line 445: Does this actually display a dropdown???

ui/page/transaction/transactions_page.go
```diff
@@ -384,15 +427,29 @@ func (pg *TransactionsPage) HandleUserInteractions() {
...
+
+	if pg.tab.Changed() {
+		if pg.isHomepageLayout {
+			pg.initWalletDropdown()
```
itswisdomagain, line 449: Why doing this when the segmented control selection is changed?

itswisdomagain (changes requested):

ui/page/transaction/transactions_page.go
```diff
@@ -101,51 +117,31 @@ func (pg *TransactionsPage) OnNavigatedTo() {
...
-						return layout.Dimensions{
-							Size: image.Point{X: selectedTabDims.Size.X, Y: tabHeight},
-						}
-					})
-				}),
-			)
-		})
-	})
+	pg.walletDropDown = pg.Theme.DropDown(items, values.WalletsDropdownGroup, 0)
+	pg.selectedWallet = pg.assetWallets[pg.walletDropDown.SelectedIndex()]
```
itswisdomagain, line 144: I just remembered something VERY important. There should not be a default selection. By default, txs for all wallets should be displayed, until and unless a user manually selects a wallet from the wallet dropdown.

This means that `pg.selectedWallet` could be nil, so always check for that before trying to use it.

ui/page/components/components.go
```diff
@@ -230,8 +230,10 @@ func TransactionTitleIcon(l *load.Load, wal sharedW.Asset, tx *sharedW.Transacti
 	return &txStatus
 }
 
-// transactionRow is a single transaction row on the transactions and overview page. It lays out a transaction's
-// direction, balance, status. isTxPage determines if the transaction should be drawn using the transactions page layout.
+// LayoutTransactionRow is a single transaction row on the transactions and overview
+// page. It lays out a transaction's direction, balance, status. isTxPage
+// determines if the transaction should be drawn using the transactions page
+// layout.
```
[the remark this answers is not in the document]
Sirmorrison: This is not my comment... I only moved it.

ui/page/components/components.go
```diff
@@ -355,10 +357,6 @@ func LayoutTransactionRow(gtx layout.Context, l *load.Load, row TransactionRow,
 								)
 							}
 
-							if isTxPage {
-								return D{}
-							}
-
 							walBalTxt := l.Theme.Label(values.TextSize12, amount)
 							walBalTxt.Color = l.Theme.Color.GrayText2
 							return walBalTxt.Layout(gtx)
```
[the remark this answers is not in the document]
Sirmorrison: If the page is a TX page, we don't need the other code executed so it better to return early.

ui/page/transaction/transactions_page.go
```diff
@@ -222,95 +222,138 @@ func (pg *TransactionsPage) loadTransactions(offset, pageSize int32) (interface{
...
 	if pg.Load.GetCurrentAppWidth() <= gtx.Dp(values.StartMobileView) {
 		return pg.layoutMobile(gtx)
 	}
 	return pg.layoutDesktop(gtx)
 }
 
-func (pg *TransactionsPage) layoutDesktop(gtx layout.Context) layout.Dimensions {
+func (pg *TransactionsPage) syncOverlay(gtx C, body layout.Widget) D {
+	// If wallet is not synced, display the overlay with the message.
+	pg.showDisabledLayout = pg.isHomepageLayout && (!pg.selectedWallet.IsSynced() || pg.selectedWallet.IsRescanning())
```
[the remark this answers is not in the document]
Sirmorrison: Its not on figma but it always applies if the page is called from the wallets page.

ui/page/transaction/transactions_page.go
```diff
@@ -222,95 +222,138 @@ func (pg *TransactionsPage) loadTransactions(offset, pageSize int32) (interface{
...
 									})
 								})
 							})
 						})
 					})
 				}),
 				layout.Expanded(func(gtx C) D {
+					if pg.isHomepageLayout && len(pg.assetWallets) > 1 {
+						return pg.walletDropDown.Layout(gtx, 0, false)
+					}
```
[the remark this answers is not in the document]
Sirmorrison: This section is just to hide the dropdown if there is only one wallet.

ui/page/transaction/transactions_page.go
Sirmorrison: Displays one dropdown at a time. So its a method the minimizes an opened dropdown and open the new one.

ui/page/transaction/transactions_page.go
Sirmorrison: The staking activities tab only applies to DCR wallets so i am using this to remove all other wallet types, leaving only DCR.

ui/page/transaction/transactions_page.go
Sirmorrison: Ermmmmm i did this earlier and the comment was to ensure it was per wallet. No sure i am going back to implementing it.

Sirmorrison: Also, to load txs, we need a wallet and currently, we can are only loading 20txs at a time. If this will be implemented, it will be in a totally different PR.

dreacot: > Ermmmmm i did this earlier and the comment was to ensure it was per wallet. No sure i am going back to implementing it.

reference the comment please.

if iirc correctly what was initially implemented was an asset dropdown and now a wallet dropdown e.g
{btc asset}
{dcr asset}

clicking on btc assets showed you transactions for all your btc wallets combined

i left a comment saying that was not correct it should be a wallet dropdown, e.g all wallets you own would show on the dropdown, so that when a wallet is selected it shows transactions for that wallet only.

>There should not be a default selection

@itswisdomagain comment is correct, there should not be a default wallet being selected on initial page visit, transactions for all wallets you own should show up, until you specifically select a wallet from the dropdown and only then do you see tx for the particular wallet selected

ui/page/components/components.go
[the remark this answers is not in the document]
itswisdomagain: But you deleted the code that returns early. That's what I'm asking about.

ui/page/transaction/transactions_page.go
```diff
@@ -54,27 +51,43 @@ type TransactionsPage struct {
...
-	txTypeDropDown   *cryptomaterial.DropDown
+	txTypeDropDown *cryptomaterial.DropDown
+	walletDropDown *cryptomaterial.DropDown
+
 	transactionList  *cryptomaterial.ClickableList
 	previousTxFilter int32
 	scroll           *components.Scroll[*sharedW.Transaction]
 
-	tabs *cryptomaterial.ClickableList
+	tab *cryptomaterial.SegmentedControl
```
itswisdomagain, line 61: Just `tab` is not a good name. What tab is this? `txCategoryTab`? Please rename.

ui/page/transaction/transactions_page.go
```diff
@@ -54,27 +51,43 @@ type TransactionsPage struct {
...
 
 	materialLoader material.LoaderStyle
+
+	sourceWalletSelector *components.WalletAndAccountSelector
```
itswisdomagain, line 65: Is there a destination wallet selector as well?

ui/page/transaction/transactions_page.go
```diff
@@ -101,51 +115,30 @@ func (pg *TransactionsPage) OnNavigatedTo() {
...
-				layout.Stacked(func(gtx C) D {
-					if i != pg.selectedTabIndex {
-						return D{}
-					}
+// initWalletSelector initializes the wallet selector dropdown to enable
+// filtering transactions for a specific wallet when this page is used to
+// display transactions for multiple wallets.
+func (pg *TransactionsPage) initWalletDropdown() {
+	pg.assetWallets = pg.WL.AllSortedWalletList()
+	if pg.tab.SelectedSegment() != values.String(values.StrTxOverview) {
```
itswisdomagain, line 123: One of the tabs is "tx overview"?

ui/page/transaction/transactions_page.go
```diff
@@ -101,51 +115,30 @@ func (pg *TransactionsPage) OnNavigatedTo() {
...
+		pg.assetWallets = pg.WL.SortedWalletList(utils.DCRWalletAsset)
+	}
 
-					tabHeight := gtx.Dp(values.MarginPadding2)
-					tabRect := image.Rect(0, 0, selectedTabDims.Size.X, tabHeight)
+	if len(pg.assetWallets) == 1 {
+		pg.selectedWallet = pg.assetWallets[0]
+	} else {
```
itswisdomagain, line 129:
```suggestion
	pg.selectedWallet = pg.assetWallets[0]
	if len(pg.assetWallets) > 1 {
```

ui/page/transaction/transactions_page.go
```diff
@@ -222,93 +219,135 @@ func (pg *TransactionsPage) loadTransactions(offset, pageSize int32) ([]*sharedW
...
+	overlay := layout.Stacked(func(gtx C) D {
+		if pg.walletNotReady() && pg.isHomepageLayout {
+			gtx = gtx.Disabled()
+			return components.DisablePageWithOverlay(pg.Load, nil, gtx, values.String(values.StrFunctionUnavailable), nil)
+		}
+		return D{}
+	})
+
+	return layout.Stack{}.Layout(gtx, layout.Expanded(body), overlay)
+}
```
itswisdomagain, line 243: 3 problems I see here.

1. This `showDisabledLayout` method doesn't always show a disabled layout. Maybe rename the method or remove it entirely.
2. If there's no need to disable the layout, no need to do `layout.Stack{}.Layout(gtx, layout.Expanded(body), overlay)` with an empty overlay. Just return the original body!
3. `components.DisablePageWithOverlay` is being abused and this PR should not follow that pattern of abuse. The function is meant to disable a page or a widget and apply an overlay over the widget. If you're not passing a page/widget, then you should not be using that function. Apparently passing a `nil` causes the fn to return the overlay for you to use as you like, let's have another function that gives you an overlay without asking you for a page or widget. It should also be a theme method, so maybe `theme.OverlayWidget(txt, btn) layout.Widget`.

ui/page/transaction/transactions_page.go
```diff
@@ -222,93 +219,135 @@ func (pg *TransactionsPage) loadTransactions(offset, pageSize int32) ([]*sharedW
...
+						return pg.walletDropDown.Layout(gtx, 0, false)
+					}
+					return D{}
+				}),
+				layout.Expanded(func(gtx C) D {
+					if pg.walletNotReady() && pg.isHomepageLayout {
+						return D{}
+					}
 					return pg.txTypeDropDown.Layout(gtx, 0, true)
 				}),
```
itswisdomagain, line 319: Instead of these additional widgets in the stack that could be empty, only display useful widgets.
Something like this should work:

```
txListWidget :=  layout.Inset{Top: values.MarginPadding60}.Layout(gtx, func(gtx C) D {
	return pg.showDisabledLayout(gtx, func(gtx C) D {...}
})

if pg.walletNotReady() && pg.walletDropDown == nil {
	return txListWidget // nothing else to display on this page at this time
}

pageElements := []layout.StackChild{layout.Expanded(txListWidget)}
if pg.walletDropDown != nil { // display wallet dropdown if available
	walletDropdownWidget := layout.Expanded(func(gtx C) D {
		return pg.walletDropDown.Layout(gtx, 0, false)
	})
	pageElements = append(pageElements, walletDropdownWidget)
}
if !pg.walletNotReady() { // display tx dropdown if selected wallet is ready
	txDropdownWidget := layout.Expanded(func(gtx C) D {
		return pg.txTypeDropDown.Layout(gtx, 0, false)
	})
	pageElements = append(pageElements, txDropdownWidget)
}

return layout.Stack{Alignment: layout.NW}.Layout(gtx, pageElements...)
```

ui/page/transaction/transactions_page.go
itswisdomagain: I mean something like this:
```
func CentralizeWidget(gtx C, widget layout.Widget) D {
	return cryptomaterial.LinearLayout{
		Width:       cryptomaterial.MatchParent,
		Height:      cryptomaterial.WrapContent,
		Orientation: layout.Horizontal,
		Direction:   layout.Center,
	}.Layout2(gtx, widget)
}
```

Then here you use as:
```
cryptomaterial.CentralizeWidget(gtx, func(gtx C) D {
	return layout.Inset{Bottom: values.MarginPadding16}.Layout(gtx, pg.tab.Layout)
})
```

itswisdomagain (changes requested):

ui/page/components/components.go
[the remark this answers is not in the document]
Sirmorrison: its currently in line 125.

ui/page/transaction/transactions_page.go
Sirmorrison: "Transaction Overview".. yes.

dreacot:
shouldn't the drop down say, all wallets if no specific wallet is selected?

and when you select a wallet you have the option to reselect `all wallets` from the dropdown to see tx fro all wallets

[Screencast from 2023-11-22 17.17.37.webm](https://github.com/crypto-power/cryptopower/assets/25265396/27af974f-2f00-4732-a8fc-7e3545692e27)

dreacot (approved):

ui/page/transaction/transactions_page.go
```diff
@@ -209,99 +276,159 @@ func (pg *TransactionsPage) loadTransactions(offset, pageSize int32) ([]*sharedW
...
+func (pg *TransactionsPage) Layout(gtx C) D {
 	if pg.Load.GetCurrentAppWidth() <= gtx.Dp(values.StartMobileView) {
 		return pg.layoutMobile(gtx)
 	}
 	return pg.layoutDesktop(gtx)
 }
 
-func (pg *TransactionsPage) layoutDesktop(gtx layout.Context) layout.Dimensions {
+func (pg *TransactionsPage) walletNotReady() bool {
+	return pg.selectedWallet != nil && (!pg.selectedWallet.IsSynced() || pg.selectedWallet.IsRescanning())
```
itswisdomagain, line 299: I think wallet is not ready if it is no wallet is selected as well. Am I correct?
So basically, wallet is not ready if selected wallet is nil or it's not synced or it's rescanning.
```suggestion
	return pg.selectedWallet == nil || !pg.selectedWallet.IsSynced() || pg.selectedWallet.IsRescanning()
```

ui/page/transaction/transactions_page.go
```diff
@@ -209,99 +276,159 @@ func (pg *TransactionsPage) loadTransactions(offset, pageSize int32) ([]*sharedW
...
+			cryptomaterial.FillMax(gtx, overlayColor, 10)
+
+			lbl := pg.Theme.Label(values.TextSize20, values.String(values.StrFunctionUnavailable))
+			lbl.Font.Weight = font.SemiBold
+			lbl.Color = pg.Theme.Color.PageNavText
+			return cryptomaterial.CentralizeWidget(gtx, lbl.Layout)
+		}))
+	}
+
+	return layout.Stack{Alignment: layout.NW}.Layout(gtx, layoutContents...)
```
itswisdomagain, line 379: What's the purpose of the NW alignment in this stack?

Also, `layoutContents` may have only one child (if there's no disabled overlay)`. Is it necessary to wrap that single child in a stack layout? I really don't think so. This should work:
```
txListWidget := func(gtx C) D {
	return layout.Inset{Top: values.MarginPadding60}.Layout(gtx, func(gtx C) D {
		...
	}
}

showOverlay := pg.walletNotReady() && pg.isHomepageLayout
if !showOverlay {
	return txListWidget(gtx)
}

return layout.Stack{Alignment: layout.NW}.Layout(gtx, 
	layout.Expanded(txListWidget),
	layout.Stacked(func(gtx C) D {
		gtx = gtx.Disabled()
		overlayColor := pg.Theme.Color.Gray3
		overlayColor.A = 220
		gtx.Constraints.Min.X = gtx.Constraints.Max.X
		gtx.Constraints.Min.Y = gtx.Constraints.Max.Y - gtx.Dp(values.MarginPadding60)
		cryptomaterial.FillMax(gtx, overlayColor, 10)

		lbl := pg.Theme.Label(values.TextSize20, values.String(values.StrFunctionUnavailable))
		lbl.Font.Weight = font.SemiBold
		lbl.Color = pg.Theme.Color.PageNavText
		return cryptomaterial.CentralizeWidget(gtx, lbl.Layout)
	})
)
```

ui/page/transaction/transactions_page.go
```diff
@@ -209,99 +276,159 @@ func (pg *TransactionsPage) loadTransactions(offset, pageSize int32) ([]*sharedW
...
-		line := pg.Theme.Separator()
-		line.Color = pg.Theme.Color.Gray3
-		items = append(items, layout.Rigid(line.Layout))
-		items = append(items, layout.Rigid(pg.sectionNavTab))
+func (pg *TransactionsPage) layoutDesktop(gtx C) D {
+	items := []layout.FlexChild{}
+	if pg.selectedWallet == nil || pg.selectedWallet.GetAssetType() == utils.DCRWalletAsset {
+		// Only show tx category navigation txCategoryTab for DCR wallets.
+		items = append(items, layout.Rigid(pg.txCategoriesNav))
+	}
```
itswisdomagain, line 387: Uhm, not really. Also show the tab if there are dcr wallets but no wallet is selected at the moment. Only hide it if there are no dcr wallets at all or the currently selected wallet is not a dcr wallet.

`pg.selectedWallet` may be nil but it doesn't mean that a dcr wallet exists and if a dcr wallet does not exist, no need to show the stake tab.

ui/page/transaction/transactions_page.go
```diff
@@ -209,99 +276,159 @@ func (pg *TransactionsPage) loadTransactions(offset, pageSize int32) ([]*sharedW
...
+	if !pg.walletNotReady() && !pg.showLoader {
+		pg.ParentWindow().Reload() //refresh UI to display updated txType dropdown
+
+		txDropdownWidget := layout.Expanded(func(gtx C) D {
+			return pg.txTypeDropDown.Layout(gtx, 0, true)
+		})
+		pageElements = append(pageElements, txDropdownWidget)
+	}
+
+	return layout.Stack{Alignment: layout.NW}.Layout(gtx, pageElements...)
```
itswisdomagain, line 428: Another scenario where there could be only one item in `pageElements` and wrapping it in a stack is unnecessary.
Also, not sure the alignment is necessary.

ui/page/transaction/transactions_page.go
```diff
@@ -209,99 +276,159 @@ func (pg *TransactionsPage) loadTransactions(offset, pageSize int32) ([]*sharedW
...
+						wallTxs := pg.scroll.FetchedData()
+						return pg.transactionList.Layout(gtx, len(wallTxs), func(gtx C, index int) D {
+							tx, wal := pg.txAndWallet(wallTxs[index])
+							return layout.Flex{Axis: layout.Vertical}.Layout(gtx,
+								layout.Rigid(func(gtx C) D {
+									isHiddenAssetsInfo := true
+									if pg.selectedWallet == nil {
+										isHiddenAssetsInfo = !isHiddenAssetsInfo
+									}
+									return components.LayoutTransactionRow(gtx, pg.Load, wal, tx, isHiddenAssetsInfo)
```
itswisdomagain, line 339: What's this boolean that's being passed to `components.LayoutTransactionRow`? The fn says it accepts a `isTxPage bool` (which I don't even know what it means) but here you're passing `isHiddenAssetsInfo` and somewhere else you pass `showAssets`. What does this boolean variable really control?

ui/page/transaction/transactions_page.go
```diff
@@ -371,35 +506,58 @@ func (pg *TransactionsPage) HandleUserInteractions() {
 		break
 	}
 
+	if pg.isHomepageLayout && pg.walletDropDown != nil && pg.walletDropDown.Changed() {
```
itswisdomagain, line 509: As long as there's a dropdown, use it. Whether it's home page layout or not. If there should only be a dropdown for homepage layout, then make sure it is nil if it is not home page layout.
```suggestion
	if pg.walletDropDown != nil && pg.walletDropDown.Changed() {
```

ui/page/transaction/transactions_page.go
```diff
@@ -371,35 +506,58 @@ func (pg *TransactionsPage) HandleUserInteractions() {
...
+
 	if clicked, selectedItem := pg.transactionList.ItemClicked(); clicked {
 		transactions := pg.scroll.FetchedData()
-		pg.ParentNavigator().Display(NewTransactionDetailsPage(pg.Load, transactions[selectedItem], false))
+		tx, wal := pg.txAndWallet(transactions[selectedItem])
+		pg.ParentNavigator().Display(NewTransactionDetailsPage(pg.Load, wal, tx, false))
+	}
+
+	dropDownList := []*cryptomaterial.DropDown{pg.txTypeDropDown}
+	if pg.isHomepageLayout && pg.walletDropDown != nil {
```
itswisdomagain, line 522:
```suggestion
	if pg.walletDropDown != nil {
```

ui/page/transaction/transactions_page.go
```diff
@@ -371,35 +506,58 @@ func (pg *TransactionsPage) HandleUserInteractions() {
...
-		pg.selectedTabIndex = clickedTabIndex
 		pg.refreshAvailableTxType()
 		go pg.scroll.FetchScrollData(false, pg.ParentWindow())
 	}
 }
 
 func (pg *TransactionsPage) listenForTxNotifications() {
+	if pg.selectedWallet == nil {
+		return
+	}
```
itswisdomagain, line 541: Listen for ntfns for all wallets. But ignore ntfns if the wallet sending the ntfn is not the currently selected wallet. Can make this a TODO for now, then I can implement it in the PR I plan to follow this up with.

itswisdomagain (changes requested):

ui/cryptomaterial/dropdown.go
```diff
@@ -152,12 +156,18 @@ func (d *DropDown) layoutActiveIcon(gtx layout.Context, index int) D {
...
+		item = d.items[itemIndex]
+	}
+
+	radius := Radius(8)
+	clickable := d.clickable
+	if d.isOpen {
+		radius = Radius(0)
+		if item.clickable != nil {
+			clickable = item.clickable
+		}
```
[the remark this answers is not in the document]
Sirmorrison: `item.Clickable` is never nil when the dropdown is opened because the item at index 0 is always selected by default. Now that we are removing the ability to set a default selected index, by setting it to -1, when dropdown is opened, `item.clickable` will be nil, which is buggy and will cause the app to crash.

ui/cryptomaterial/dropdown.go
```diff
@@ -152,12 +156,18 @@ func (d *DropDown) layoutActiveIcon(gtx layout.Context, index int) D {
...
+	var item DropDownItem
+	if itemIndex > -1 {
+		item = d.items[itemIndex]
+	}
+
+	radius := Radius(8)
+	clickable := d.clickable
+	if d.isOpen {
+		radius = Radius(0)
+		if item.clickable != nil {
```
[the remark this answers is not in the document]
Sirmorrison: All Possible areas where items could panic was addressed.

ui/page/transaction/transactions_page.go
Sirmorrison: What's the purpose of the NW alignment in this stack?: Its used to align the stacked widget.

ui/page/transaction/transactions_page.go
Sirmorrison: You are not correct.. the fact that a wallet is not selected does not mean we should not show txs of wallet that are synced.

When fetching txs for multi wallets, we check the sync status.. if it not synced, we skip that wallet.

ui/page/transaction/transactions_page.go
Sirmorrison: The bool is an indicator for the tx list to show some additional information about the tx.. example the wallet it belongs to.

ui/page/transaction/transactions_page.go
```diff
@@ -147,54 +147,121 @@ func (pg *TransactionsPage) pageTitle(gtx C) D {
...
-			item := cryptomaterial.DropDownItem{}
-			if pg.selectedTabIndex == 0 {
-				item.Text = fmt.Sprintf("%s (%d)", name, countfn(fieldtype))
-			} else {
-				item.Text = name
+			items := []cryptomaterial.DropDownItem{}
+			mapinfo, keysinfo := components.TxPageDropDownFields(pg.getAssetType(), pg.selectedTxCategoryTab)
+			for _, name := range keysinfo {
+				fieldtype := mapinfo[name]
+				items = append(items, cryptomaterial.DropDownItem{
```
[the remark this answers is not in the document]
Sirmorrison: This was resolved.

Sirmorrison:
> shouldn't the drop down say, all wallets if no specific wallet is selected?
> 
> and when you select a wallet you have the option to reselect `all wallets` from the dropdown to see tx fro all wallets
> 
>  Screencast.from.2023-11-22.17.17.37.webm

About this, we are currently using all sorted assets. Adding all, as a dropdown option is a PITA as we will need to start checking a couple of dropdown index and hardcoding params.

We we currently have is.. if there is no selected wallet, we display all the txs when we select a wallet, its sorted until the user switches tab.

dreacot:
> > shouldn't the drop down say, all wallets if no specific wallet is selected?
> > and when you select a wallet you have the option to reselect `all wallets` from the dropdown to see tx fro all wallets
> > Screencast.from.2023-11-22.17.17.37.webm
> 
> About this, we are currently using all sorted assets. Adding all, as a dropdown option is a PITA as we will need to start checking a couple of dropdown index and hardcoding params.
> 
> We we currently have is.. if there is no selected wallet, we display all the txs when we select a wallet, its sorted until the user switches tab.

even if we might not make that change in this PR, we still have to consider it, because ideally that's how it should work right?

ui/page/transaction/transactions_page.go
```diff
@@ -147,163 +146,271 @@ func (pg *TransactionsPage) pageTitle(gtx C) D {
...
+	if wal == nil {
+		txs, totalTxs, err = pg.multiWalletTxns(offset, pageSize)
+	} else {
+		txs, totalTxs, err = pg.loadTransactions(wal, offset, pageSize)
+	}
+
+	isReset = pg.previousTxFilter != pg.txFilter
+	if isReset {
+		// reset the offset to zero
+		offset = 0
```
dreacot, line 208: This is probably where you lint error is coming from, why is offset set to zero here, it servers no purpose

Sirmorrison:
follow up PR https://github.com/crypto-power/cryptopower/pull/268

ui/cryptomaterial/dropdown.go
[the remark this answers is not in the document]
itswisdomagain: Are you basically saying that `item` cannot be nil if `d.isOpen`?
Because in line 159-162 above, we only give `item` a value if `itemIndex > -1`. So if itemIndex is -1, item will be nil.

And this point where you do `if item.clickable != nil` will panic. Unless `d.isOpen` is only true if `itemIndex > -1`.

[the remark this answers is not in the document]
itswisdomagain: > All Possible areas where items could panic was addressed.

I don't see where this is done. Point me to it.

ui/cryptomaterial/dropdown.go
```diff
@@ -193,11 +203,14 @@ func (d *DropDown) layoutOption(gtx layout.Context, itemIndex int) D {
...
 				Left:  unit.Dp(5),
 			}.Layout(gtx, func(gtx C) D {
-				lbl := d.theme.Body2(item.Text)
-				if !d.isOpen && len(item.Text) > 14 {
-					lbl.Text = item.Text[:14] + "..."
+				txt := item.Text
+				if !d.isOpen && len(txt) > 14 {
+					txt = item.Text[:14] + "..."
+				} else if txt == "" {
+					txt = "Select a wallet"
```
itswisdomagain, line 210: No. You can pass a `prompt string` but don't hard code this for all dropdowns that have no txt.

ui/page/components/components.go
```diff
@@ -226,9 +226,12 @@ func TransactionTitleIcon(l *load.Load, wal sharedW.Asset, tx *sharedW.Transacti
...
 
-// transactionRow is a single transaction row on the transactions and overview page. It lays out a transaction's
-// direction, balance, status. isTxPage determines if the transaction should be drawn using the transactions page layout.
-func LayoutTransactionRow(gtx layout.Context, l *load.Load, wal sharedW.Asset, tx *sharedW.Transaction, isTxPage bool) layout.Dimensions {
+// LayoutTransactionRow is a single transaction row on the transactions and overview
+// page. It lays out a transaction's direction, balance, status. isHiddenTxAssetInfo
+// determines if the transaction should display additional information about the tx
+// such as the wallet the tx belong to etc. This is usefil on pages where
+// the tx is displayed from multi wallets.
+func LayoutTransactionRow(gtx layout.Context, l *load.Load, wal sharedW.Asset, tx *sharedW.Transaction, isHiddenTxAssetInfo bool) layout.Dimensions {
```
itswisdomagain, line 234:
```suggestion
func LayoutTransactionRow(gtx layout.Context, l *load.Load, wal sharedW.Asset, tx *sharedW.Transaction, hideTxAssetInfo bool) layout.Dimensions {
```

ui/page/components/components.go
```diff
@@ -628,7 +628,7 @@ func TimeFormat(secs int, long bool) string {
 
 // TxPageDropDownFields returns the fields for the required drop down with the
 // transactions view page. Since maps access of items order is always random
-// an array of keys is provided guarrantee the dropdown order will always be
+// an array of keys is provided guarantee the dropdown order will always be
```
itswisdomagain, line 631:
```suggestion
// an array of keys is provided to guarantee the dropdown order will always be
```

ui/page/transaction/transactions_page.go
```diff
@@ -47,34 +50,53 @@ type TransactionsPage struct {
...
 
 	materialLoader material.LoaderStyle
+
+	walletSelector *components.WalletAndAccountSelector
+	assetWallets   []sharedW.Asset
+	selectedWallet sharedW.Asset
+
+	isHomepageLayout,
+	showLoader,
+	DCRWalletExist bool
```
itswisdomagain, line 73:
```suggestion
	dcrWalletExists bool
```

ui/page/transaction/transactions_page.go
```diff
@@ -47,34 +50,53 @@ type TransactionsPage struct {
...
-	pg.tabs = l.Theme.NewClickableList(layout.Horizontal)
-	pg.tabs.IsHoverable = false
+	pg.scroll = components.NewScroll(l, pageSize, pg.fetchTransactions)
 
 	pg.transactionList.Radius = cryptomaterial.Radius(14)
 	pg.transactionList.IsShadowEnabled = true
 
-	pg.materialLoader = material.Loader(l.Theme.Base)
+	pg.materialLoader = material.Loader(pg.Theme.Base)
+	go pg.scroll.FetchScrollData(false, pg.ParentWindow())
```
itswisdomagain, line 99: This should happen in OnNavigatedTo. When the page is initialized, it is not guaranteed to be displayed immediately. Data fetching should happen when the page is about to be displayed, not when it is initialized.

ui/page/transaction/transactions_page.go
```diff
@@ -85,60 +107,37 @@ func NewTransactionsPage(l *load.Load) *TransactionsPage {
...
-						}
-					})
-				}),
-			)
-		})
-	})
+// initWalletSelector initializes the wallet selector dropdown to enable
+// filtering transactions for a specific wallet when this page is used to
+// display transactions for multiple wallets.
+func (pg *TransactionsPage) initWalletDropdown() {
```
itswisdomagain, line 117: initWalletSelector or initWalletDropdown?

ui/page/transaction/transactions_page.go
```diff
@@ -147,163 +146,272 @@ func (pg *TransactionsPage) pageTitle(gtx C) D {
...
 	}
-
 	pg.txTypeDropDown = pg.Theme.DropDown(items, values.TxDropdownGroup, 2)
 
-	go func() {
-		countfn := func(fType int32) int {
-			count, _ := wal.CountTransactions(fType)
-			return count
+	// only show tx count for regular txs, not staking
+	if pg.selectedTxCategoryTab == 0 {
```
itswisdomagain, line 165: A few lines before this, I saw `pg.txCategoryTab.SelectedSegment() != values.String(values.StrTxOverview)`. That's easier to understand. Please remove the `pg.selectedTxCategoryTab` variable.
```suggestion
	if pg.txCategoryTab.SelectedSegment() == values.String(values.StrTxOverview) {
```

itswisdomagain (changes requested):

ui/page/components/items_scroll.go
```diff
@@ -68,6 +68,10 @@ func (s *Scroll[T]) FetchScrollData(isReverse bool, window app.WindowNavigator)
 	}
 	s.mu.Unlock()
 
+	// set isReset  to true here since all the callers of FetchScrollData()
+	// needs the data to be refreashed and data reset.
+	// TODO: set isreverse to default false all the callers of this
+	// FetchScrollData() is not initiating a reverse scroll action.
 	s.fetchScrollData(isReverse, window)
```
itswisdomagain, line 75: This comment isn't clear. `set isReset  to true here`, what/where is `isReset`?
Since there's a follow up PR working on this file, let's keep changes here minimal and only absolutely necessary to reduce conflict and make this PR easier to follow.

Actually, I think undo all the changes to this file, including this comment and the `resetList` method name change.

ui/page/transaction/transactions_page.go
```diff
@@ -47,34 +50,52 @@ type TransactionsPage struct {
...
 	previousTxFilter int32
-	scroll           *components.Scroll[*sharedW.Transaction]
+	scroll *components.Scroll[*multiWalletTx]
 
-	tabs *cryptomaterial.ClickableList
+	txCategoryTab *cryptomaterial.SegmentedControl
 
 	materialLoader material.LoaderStyle
+
+	walletSelector *components.WalletAndAccountSelector
```
itswisdomagain, line 67: Not sure this is used.

itswisdomagain (approved):
I'm yet to fully test but the code changes look okay now. Just these 2 comments.

dreacot (approved):

PR #208, opened 2023-10-27
Multi: Create DEX main page

ukane-philemon:
Closes #187
 
The new design(for the order form) in this PR has been agreed upon by @itswisdomagain, everything else stays the same.

Also, some notable UI elements have been omitted in this PR (e.g the order filter row).

DCRDEXscreen_3

PR #238, opened 2023-11-04
Integrate DCRDEX for DEX onboarding page

ukane-philemon:
closes https://github.com/crypto-power/cryptopower/issues/209

Depends on #208 

Note for reviewers:
1. This MR does not implement the required methods to make our existing assets compatible with dex core wallet interface https://github.com/decred/dcrdex/blob/master/client/asset/interface.go. That will be achieved in a separate PR.
2. Actual bond posting depends on 1 above.


ui/window.go
```diff
@@ -155,9 +158,30 @@ func (win *Window) NewLoad() (*load.Load, error) {
...
 }
 
+// RunDexCore starts a goroutine that starts/initializes dex core. It blocks
+// until dex core is ready.
+func (win *Window) RunDexCore(ctx context.Context) {
+	go func() {
+		win.load.DexCore.Run(ctx)
+		win.Quit <- struct{}{} // in the event that Run returns prematurely prior to context cancellation
+	}()
+	<-win.load.DexCore.Ready()
```
itswisdomagain, line 182: This will have more use in the DEX page, can show a loader while DEX is initializing and remove the loader when dexCore is ready.

dexcore/core.go
```diff
@@ -0,0 +1,92 @@
...
+	InitializeClient(pw, seed []byte) error
+	DiscoverAccount(dexAddr string, pass []byte, certI any) (*core.Exchange, bool, error)
+	BondsFeeBuffer(assetID uint32) uint64
+	PostBond(form *core.PostBondForm) (*core.PostBondResult, error)
+	NotificationFeed() <-chan core.Notification
+	Login(pw []byte) error
+}
+
+// New returns a *DEXCore that satisfies Core.
+func New(root, lang string, net libutils.NetworkType) (*DEXCore, error) {
```
itswisdomagain, line 27: This file, above this New func, will be the right place to define the DEXCore type.

dexcore/core.go
```diff
@@ -0,0 +1,92 @@
...
+	Run(ctx context.Context)
+	Ready() <-chan struct{}
+	IsInitialized() bool
+	InitializeClient(pw, seed []byte) error
+	DiscoverAccount(dexAddr string, pass []byte, certI any) (*core.Exchange, bool, error)
+	BondsFeeBuffer(assetID uint32) uint64
+	PostBond(form *core.PostBondForm) (*core.PostBondResult, error)
+	NotificationFeed() <-chan core.Notification
+	Login(pw []byte) error
+}
```
itswisdomagain, line 24: This should be defined by consumers, I think. Similar to how dcrdex/client/webserver defines a Core interface for methods that are needed by that package.

dexcore/core.go
```diff
@@ -0,0 +1,92 @@
...
+
+// New returns a *DEXCore that satisfies Core.
+func New(root, lang string, net libutils.NetworkType) (*DEXCore, error) {
+	dexNet := toDEXNet(string(net))
```
itswisdomagain, line 28: I'll prefer `dexNet := string(net); if dexNet == "simulation" or "regression", dexNet = "simnet"`.

dexcore/core.go
```diff
@@ -0,0 +1,92 @@
...
+	network, err := dex.NetFromString(dexNet)
+	if err != nil {
+		return nil, fmt.Errorf("dex.NetFromString error: %w", err)
+	}
+
+	dbPath := filepath.Join(root, string(net), "dexc.db")
```
itswisdomagain, line 34: I have a guess why you're using string(net) instead of dexNet, but add a comment.

dexcore/core.go
```diff
@@ -0,0 +1,92 @@
...
+		Net:                network,
+		Logger:             log,
+		Language:           lang,
+		UnlockCoinsOnLogin: true,
+		NoAutoWalletLock:   true,
+	}
+
+	clientCore, err := core.New(cfg)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to initialize dex core: %w", err)
```
itswisdomagain, line 46:
```suggestion
		return nil, fmt.Errorf("failed to initialize dex core: %w", err)
```

dexcore/core.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+
+	dc := &DEXCore{
+		Core:            clientCore,
+		bondBufferCache: make(map[uint32]valStamp),
+	}
+
+	return dc, err
```
itswisdomagain, line 54:
```suggestion
	return &DEXCore{
		Core:            clientCore,
		bondBufferCache: make(map[uint32]valStamp),
	}, nil
```

dexcore/core.go
```diff
@@ -0,0 +1,92 @@
...
+		return nil, fmt.Errorf("dex.NetFromString error: %w", err)
+	}
+
+	dbPath := filepath.Join(root, string(net), "dexc.db")
+	cfg := &core.Config{
+		DBPath:             dbPath,
+		Net:                network,
+		Logger:             log,
+		Language:           lang,
+		UnlockCoinsOnLogin: true,
```
itswisdomagain, line 40: Why? Won't this unlock utxos that were locked by some other process for some other reason? E.g. during a send or ticket purchase.

dexcore/core.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+
+	return dc, err
+}
+
+// BondsFeeBuffer is a caching helper for the bonds fee buffer. Values for a
+// given asset are cached for 45 minutes. These values are meant to provide a
+// sensible but well-padded fee buffer for bond transactions now and well into
+// the future, so a long expiry is appropriate.
+func (dc *DEXCore) BondsFeeBuffer(assetID uint32) (feeBuffer uint64) {
```
itswisdomagain, line 61: Why do we need a cache. Don't see that in the comment.

itswisdomagain (changes requested):

wallet/wallet.go
```diff
@@ -29,21 +29,16 @@ type Wallet struct {
...
-func NewWallet(root, net, version, logFolder string, buildDate time.Time) (*Wallet, error) {
+func NewWallet(root string, net libutils.NetworkType, version, logFolder string, buildDate time.Time) (*Wallet, error) {
 	if root == "" {
 		return nil, fmt.Errorf("root directory cannot be empty")
 	}
 
-	resolvedNetType := libutils.ToNetworkType(net)
-	if resolvedNetType == libutils.Unknown {
-		return nil, fmt.Errorf("network type is not supportted: %s", net)
-	}
```
ukane-philemon, line 42: @itswisdomagain re https://github.com/crypto-power/cryptopower/pull/238#discussion_r1382516892, The "testnet3" string was corrected here(libutils.ToNetworkType(net) will return "testnet" for "testnet3") so I'm sure we are on the right part.

config.go
```diff
@@ -230,33 +226,39 @@ func loadConfig() (*config, error) {
 	if cfg.MaxLogZips < 0 {
 		cfg.MaxLogZips = 0
 	}
-	net := cfg.Network
-	if net == "testnet" {
-		net = "testnet3"
-	}
-	logDir := filepath.Join(cfg.LogDir, net)
```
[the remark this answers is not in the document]
ukane-philemon: >  but check for other possible effects of this change

Only the log dir was affected by this change.

libwallet/utils/config.go
```diff
@@ -88,7 +90,7 @@ func (str AssetType) ToFull() string {
 	case LTCWalletAsset:
 		return "Litecoin"
 	default:
-		return "Unknown"
+		return string(str)
```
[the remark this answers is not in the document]
ukane-philemon: We will likely use the `libwallet/utils.AssetType` type when dealing with dex assets, most of these assets are not supported by crypto power but we will show these markets to users.

main.go
```diff
@@ -23,7 +23,9 @@ var (
 	// BuildDate is the date the application was built. It is set using the -ldflags
 	BuildDate string
 	// BuildEnv is the build environment. It is set using the -ldflags
-	BuildEnv = wallet.DevBuild
+	BuildEnv       = wallet.DevBuild
+
+	appCtx, cancel = context.WithCancel(context.Background())
```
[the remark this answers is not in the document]
ukane-philemon: No particular reason. Only dex core uses it.

main.go
```diff
@@ -81,12 +74,8 @@ func main() {
 	}
 
 	// if debuglevel is passed at commandLine persist the option.
-	if cfg.DebugLevel != "" && wal.GetAssetsManager().IsAssetManagerDB() {
-		wal.GetAssetsManager().SetLogLevels(cfg.DebugLevel)
-	}
-
 	if wal.GetAssetsManager().IsAssetManagerDB() {
-		// now that assets manager is up, set stored debuglevel
+		wal.GetAssetsManager().SetLogLevels(cfg.DebugLevel)
```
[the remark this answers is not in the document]
ukane-philemon: > the intention is to use the last stored log level rather than default.

Hmm, I agree. I'll revert and improve the comment.

dexcore/core.go
ukane-philemon: The reason is the first comment in the method, moved it though.

ui/page/dcrdex/dcrdex_page.go
```diff
@@ -63,24 +75,81 @@ func (pg *DEXPage) OnNavigatedTo() {
...
 			}.Layout(gtx,
-				layout.Flexed(1, pg.CurrentPage().Layout),
+				layout.Flexed(1, func(gtx layout.Context) layout.Dimensions {
+					if !dexSet {
+						return D{}
+					}
+					return pg.CurrentPage().Layout(gtx)
+				}),
 			)
 		}),
```
itswisdomagain, line 116: This can be defined as `pageUI` or something and if there's no need for an overlay, return this directly, else return this as part of a layout.Stack{}.

ui/page/dcrdex/dcrdex_page.go
```diff
@@ -63,24 +75,81 @@ func (pg *DEXPage) OnNavigatedTo() {
...
 			return cryptomaterial.LinearLayout{
 				Width:       cryptomaterial.MatchParent,
 				Height:      cryptomaterial.MatchParent,
 				Orientation: layout.Vertical,
 			}.Layout(gtx,
-				layout.Flexed(1, pg.CurrentPage().Layout),
+				layout.Flexed(1, func(gtx layout.Context) layout.Dimensions {
+					if !dexSet {
+						return D{}
+					}
```
itswisdomagain, line 112: This would mean having a stack layout holding an expanded layout, holding a linearlayout, holding a flexed layout just to display nothing. Please, no.

itswisdomagain (changes requested):
Taking a pause, to go again after these.

dexcore/core.go
```diff
@@ -0,0 +1,115 @@
...
+	logCloser := func() {
+		r.Close()
+	}
+
+	l, err := dex.NewLoggerMaker(r, lvl, true /* TODO: Make configurable */)
+	if err != nil {
+		return nil, nil, fmt.Errorf("Failed to initialize log: %w", err)
+	}
+
+	return l.NewLogger("DCRDEX", l.DefaultLevel), logCloser, nil
```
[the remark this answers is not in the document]
ukane-philemon: Yes, this is redundant.

ui/page/dcrdex/dcrdex_page.go
```diff
@@ -21,27 +26,34 @@ type DEXPage struct {
 
 	*load.Load
 
-	ctx       context.Context // page context
-	ctxCancel context.CancelFunc
+	dexCore Core
+
+	ctx                context.Context // page context
+	ctxCancel          context.CancelFunc
```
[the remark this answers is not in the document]
ukane-philemon: yes, we need them to exit core properly in the event that the application is shutdown.

ui/page/dcrdex/dcrdex_page.go
```diff
@@ -63,24 +75,81 @@ func (pg *DEXPage) OnNavigatedTo() {
 // eventually drawn on screen.
 // Part of the load.Page interface.
 func (pg *DEXPage) Layout(gtx C) D {
+	noMultipleWalMsg := pg.isMultipleAssetTypeWalletAvailable()
+	dexDisableMsg := pg.dexEnabled()
+	overlay := layout.Stacked(func(gtx C) D { return D{} })
+	c, _ := pg.dexCore.(*dexcore.DEXCore)
+	dexSet := c != nil
```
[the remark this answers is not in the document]
ukane-philemon: pg.dexCore is an Interface and not "nil" even if there's no underlying value. I didn't want to use `*dexcore.DEXCore` everywhere since we only need some core methods.

main.go
```diff
@@ -64,21 +65,15 @@ func main() {
...
-	logDir := filepath.Join(cfg.LogDir, net)
-	assetsManager, err := libwallet.NewAssetsManager(cfg.HomeDir, logDir, net)
+	logDir := filepath.Join(cfg.LogDir, string(cfg.net))
+	assetsManager, err := libwallet.NewAssetsManager(cfg.HomeDir, logDir, cfg.net)
 	if err != nil {
 		log.Errorf("init assetsManager error: %v", err)
 		return
 	}
 
+	assetsManager.MaxLogZips = cfg.MaxLogZips
```
itswisdomagain, line 75: This is only used in UI, not by assetsmanager. Put it in load, for now. I'm not sure the `AppInfo` struct is in master yet, but if it is, that's a better place to put it.

ui/page/dcrdex/dcrdex_page.go
```diff
@@ -21,66 +25,97 @@ type DEXPage struct {
...
+	hasMultipleWallets := pg.isMultipleAssetTypeWalletAvailable()
+	privacyModeOff := pg.WL.AssetsManager.IsHTTPAPIPrivacyModeOff(utils.ExchangeHTTPAPI)
+	var msg string
+	var actionBtn *cryptomaterial.Button
+	if !hasMultipleWallets {
+		msg = values.String(values.StrMultipleAssetTypeRequiredToTradeDEX)
+	} else if !privacyModeOff {
+		actionBtn = &pg.generalSettingsBtn
+		msg = values.StringF(values.StrNotAllowed, values.String(values.StrExchange))
+	} else if dexCore, _ := pg.dexCore.(*dexcore.DEXCore); dexCore == nil {
```
itswisdomagain, line 80: Why not just `if pg.dexCore == nil`.

ui/page/dcrdex/dcrdex_page.go
```diff
@@ -21,66 +25,97 @@ type DEXPage struct {
...
+
+	if msg != "" {
+		return components.DisablePageWithOverlay(pg.Load, nil, gtx, msg, actionBtn)
+	}
+
+	return cryptomaterial.LinearLayout{
+		Width:       cryptomaterial.MatchParent,
+		Height:      cryptomaterial.MatchParent,
+		Orientation: layout.Vertical,
+	}.Layout2(gtx, pg.CurrentPage().Layout)
```
itswisdomagain, line 92: Please, what does `Layout2` do and how is it different from `Layout`?

Also, why is pg.CurrentPage() being wrapped in a LinearLayout? Any `pg.CurrentPage()` that needs its widgets to be wrapped in this way should do the wrapping themselves, this should probably just be `pg.CurrentPage().Layout(gtx)`, unless I'm missing something.

ui/page/root/home_page.go
```diff
@@ -203,7 +212,8 @@ func (hp *HomePage) HandleUserInteractions() {
 		case values.String(values.StrWallets):
 			pg = hp.walletSelectorPage
 		case values.String(values.StrTrade):
-			pg = NewTradePage(hp.Load)
+			hp.ensureDEXIsReady()
+			pg = NewTradePage(hp.Load, hp.dexCore)
```
itswisdomagain, line 216: `hp.ensureDEXIsReady` doesn't really ensure dex is ready, dexCore might still be nil after that method returns. Should check here and NOT naviagate to the dex page if dexCore could not be initialized. Show a toast message or something instead.

ui/page/root/home_page.go
```diff
@@ -276,7 +286,8 @@ func (hp *HomePage) HandleUserInteractions() {
 			case values.String(values.StrWallets):
 				pg = hp.walletSelectorPage
 			case values.String(values.StrTrade):
-				pg = NewTradePage(hp.Load)
+				hp.ensureDEXIsReady()
+				pg = NewTradePage(hp.Load, hp.dexCore)
```
itswisdomagain, line 290: This is repeated. Let's have a method that takes item.Title and returns the page to navigate to.

ui/page/root/home_page.go
```diff
@@ -810,3 +821,33 @@ func (hp *HomePage) CalculateAssetsUSDBalance() {
 		hp.ParentWindow().Reload()
 	}
 }
+
+// initializeDEX will initialize hp.dexCore if it has not been initialized
+// before. It's a no-op if hp.dexCore has been initialized.
+func (hp *HomePage) ensureDEXIsReady() {
+	if hp.dexCore != nil && hp.dexCoreRunning {
```
itswisdomagain, line 828: Looks like just `hp.dexCore != nil` should be enough. Instead of setting `hp.dexCoreRunning` to false later, can set `hp.dexCore` to nil.

ui/page/root/home_page.go
```diff
@@ -810,3 +821,33 @@ func (hp *HomePage) CalculateAssetsUSDBalance() {
...
+		return // nothing to do
+	}
+
+	logger, logCloser, err := dexcore.NewDexLogger(filepath.Dir(hp.WL.AssetsManager.LogFile()), hp.WL.AssetsManager.GetLogLevels(), hp.WL.AssetsManager.MaxLogZips)
+	if err != nil {
+		log.Errorf("error initializing dex log: %v", err)
+		return
+	}
+
+	hp.dexCore, err = dexcore.New(hp.WL.AssetsManager.NetworkRootDir(), hp.WL.AssetsManager.GetLanguagePreference(), hp.WL.AssetsManager.NetType(), logger) // TODO: We should be able to update dex core's language when the user changes language.
```
itswisdomagain, line 838: Also put this TODO is the settings code where language is updated.

ui/page/root/home_page.go
```diff
@@ -810,3 +821,33 @@ func (hp *HomePage) CalculateAssetsUSDBalance() {
...
+// initializeDEX will initialize hp.dexCore if it has not been initialized
+// before. It's a no-op if hp.dexCore has been initialized.
+func (hp *HomePage) ensureDEXIsReady() {
+	if hp.dexCore != nil && hp.dexCoreRunning {
+		return // nothing to do
+	}
+
+	logger, logCloser, err := dexcore.NewDexLogger(filepath.Dir(hp.WL.AssetsManager.LogFile()), hp.WL.AssetsManager.GetLogLevels(), hp.WL.AssetsManager.MaxLogZips)
+	if err != nil {
+		log.Errorf("error initializing dex log: %v", err)
```
itswisdomagain, line 834:
```suggestion
		log.Errorf("error initializing dex logger: %v", err)
```

ui/page/root/home_page.go
```diff
@@ -810,3 +821,33 @@ func (hp *HomePage) CalculateAssetsUSDBalance() {
...
 }
+
+// initializeDEX will initialize hp.dexCore if it has not been initialized
+// before. It's a no-op if hp.dexCore has been initialized.
+func (hp *HomePage) ensureDEXIsReady() {
+	if hp.dexCore != nil && hp.dexCoreRunning {
+		return // nothing to do
+	}
+
+	logger, logCloser, err := dexcore.NewDexLogger(filepath.Dir(hp.WL.AssetsManager.LogFile()), hp.WL.AssetsManager.GetLogLevels(), hp.WL.AssetsManager.MaxLogZips)
```
itswisdomagain, line 832: Isn't `AssetsManager.LogFile()` currently in use by another log rotator?

ui/page/root/home_page.go
```diff
@@ -810,3 +821,33 @@ func (hp *HomePage) CalculateAssetsUSDBalance() {
...
+		return
+	}
+
+	hp.dexCoreRunning = true
+	// Use a goroutine to start dex core as it'll blocks until dex core exits.
+	go func() {
+		// TODO: We should be able to shutdown core properly if Homepage is ever
+		// destroyed and recreated while the app is still running.
+		hp.dexCore.Run(hp.dexCtx)
+		hp.dexCoreRunning = false
```
itswisdomagain, line 850: Use a page ctx as mentioned in a previous comment.
Use a `wg` as well, increment before starting the goroutine, decrement inside the goroutine after dexCore.Run(.) returns.
In OnNavigatedFrom(), cancel the page ctx, and wait on the `wg`.

itswisdomagain: It actually looks like all these can go to the dexcore package. A `StartDEXClient()` function that does all this and returns the DEXClient. Which reminds me, DEXCore should probably be DEXClient. `Core` is a dcrdex concept. In cryptopower, we're dealing with a DEXClient which uses dcrdex.Core under the hood.

ui/page/root/trade_page.go
```diff
@@ -47,6 +49,7 @@ func NewTradePage(l *load.Load) *TradePage {
 				Alignment: layout.Middle,
 			},
 		},
+		dexCore: dexCore,
```
itswisdomagain, line 52: Assign this after MasterPage is assigned.

itswisdomagain (changes requested):

libwallet/assets_manager.go
```diff
@@ -110,33 +112,28 @@ func initializeAssetsFields(rootDir, dbDriver, logDir string, netType utils.Netw
...
-	netType := utils.ToNetworkType(net)
-	if netType == utils.Unknown {
-		return nil, fmt.Errorf("network type is not supportted: %s", net)
-	}
-
 	// Create a root dir that has the path up the network folder.
-	rootDir = filepath.Join(rootDir, string(netType))
-	if err := os.MkdirAll(rootDir, utils.UserFilePerm); err != nil {
+	networkRootDir := filepath.Join(rootDir, string(netType))
+	if err := os.MkdirAll(networkRootDir, utils.UserFilePerm); err != nil {
```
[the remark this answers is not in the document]
ukane-philemon: The "root" dir might be requested by others like "dexcore" where we also append the network dir for dexcore's db. Consumers do not know that the "rootDir" is actually the "rooDir/network" and might be confusing to use app-wide, especially if we want a file in the actual "rootDir" and not "rootDir/network".

main.go
ukane-philemon: Okay.

ui/page/dcrdex/dcrdex_page.go
```diff
@@ -21,66 +25,97 @@ type DEXPage struct {
 
 	*load.Load
 
-	ctx       context.Context // page context
-	ctxCancel context.CancelFunc
+	dexCore Core
+	inited  bool // TODO: Set value
```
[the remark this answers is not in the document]
ukane-philemon: Uhmm, actually this is handy when the user is already connected/registered with their first DEX server so we don't have to display the onboarding page.

ui/page/dcrdex/dcrdex_page.go
ukane-philemon: You're okay with having `pg.dexCore  *dexcore.DEXCore`? And ditching the interface?

ui/page/dcrdex/dcrdex_page.go
ukane-philemon: > this should probably just be pg.CurrentPage().Layout(gtx)

I thought about this but wanted a way to ensure all child elements are displayed vertically. But I think we can return `pg.CurrentPage().Layout(gtx)` as you've suggested, all pages should proly handle orientation themselves.

ui/page/root/home_page.go
ukane-philemon: Okay. I think will need a `mtx` to protect hp.dexCore.

ui/page/root/home_page.go
ukane-philemon: We need just the dir `filepath.Dir(hp.WL.AssetsManager.LogFile())`.

run_tests.sh
```diff
@@ -16,6 +16,5 @@ for i in $(find . -name go.mod -type f -print); do
       --enable unconvert \
       --enable ineffassign \
       --enable goimports \
-      --enable misspell \
```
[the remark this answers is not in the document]
ukane-philemon: > We don't care about this? 

Lint errors, I think we need the PR to be clean before merge, no?

ui/page/root/home_page.go
```diff
@@ -32,6 +35,10 @@ type HomePage struct {
 
 	*load.Load
 
+	dexCtx         context.Context
```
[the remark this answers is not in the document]
ukane-philemon: The Homepage can continue to exist in the stack for the duration of the app's lifetime but its `OnNavigatedTo` and  `OnNavigatedFrom` methods are called when the user navigates to and from the `Homepage`. Using the page's `ctx` will require us to shut down and restart dexcore multiple times during the app's lifetime.

ui/page/root/home_page.go
ukane-philemon: > and NOT naviagate to the dex page if dexCore could not be initialized. Show a toast message or something instead.

We have both the CEX page and DEX page under the Trade page. This means we have to display the trade page but handle a possible nil dex client in the DEX page. I've done this in my lastest review changes.

ui/page/dcrdex/dex_onboarding_page.go
```diff
@@ -170,7 +199,24 @@ func NewDEXOnboarding(l *load.Load) *DEXOnboarding {
...
 // Part of the load.Page interface.
 func (pg *DEXOnboarding) OnNavigatedTo() {
-	pg.showLoader = false
+	ch := pg.dexc.NotificationFeed()
+	// TODO: return note feed?
+	go func() {
+		for {
+			n, ok := <-ch
+			if !ok {
+				return
```
itswisdomagain, line 208: This may also be a good place to return the note and set the channel to nil so it is not returned in OnNavigatedFrom().

ui/page/dcrdex/dex_onboarding_page.go
```diff
@@ -186,7 +232,10 @@ func (pg *DEXOnboarding) OnNavigatedFrom() {}
 // to be eventually drawn on screen.
 // Part of the load.Page interface.
 func (pg *DEXOnboarding) Layout(gtx C) D {
-	r := 8
+	return pg.pageLayout(gtx)
+}
+
+func (pg *DEXOnboarding) pageLayout(gtx C) D {
```
itswisdomagain, line 238: Merge? Only called here.

ui/page/dcrdex/dex_onboarding_page.go
```diff
@@ -450,19 +494,21 @@ func (pg *DEXOnboarding) subStepAddServer(gtx C) D {
...
 		}),
-		layout.Rigid(func(gtx C) D {
-			return pg.formFooterButtons(gtx)
-		}),
+		layout.Rigid(pg.formFooterButtons),
 	)
 }
 
 // formFooterButtons is a convenience function that prepares the required
 // buttons for each form page.
```
itswisdomagain, line 502: for each form page footer?

ui/page/dcrdex/dex_onboarding_page.go
```diff
@@ -450,19 +494,21 @@ func (pg *DEXOnboarding) subStepAddServer(gtx C) D {
...
 func (pg *DEXOnboarding) formFooterButtons(gtx C) D {
-	var addBackBtn bool
+	addBackBtn, nextBtnEnabled, backBtnEnabled := true, true, true
 	switch pg.currentStep {
-	case onboardingPostBond, onboardingChooseServer, onBoardingStepAddServer:
-		addBackBtn = true
+	case onboardingSetPassword:
+		addBackBtn = false
+	case onboardingChooseServer, onBoardingStepAddServer:
+		backBtnEnabled = !pg.dexc.IsInitialized()
```
itswisdomagain, line 509: I can't go back if I have initialized and I'm on the choose/add server step? I'm not really getting this part.

ui/page/dcrdex/dex_onboarding_page.go
```diff
@@ -494,9 +540,23 @@ func (pg *DEXOnboarding) formFooterButtons(gtx C) D {
...
+			if pg.isLoading {
+				return layout.Flex{}.Layout(gtx,
+					layout.Flexed(1, func(gtx C) D {
+						return layout.Center.Layout(gtx, func(gtx C) D {
+							gtx.Constraints.Max.X = gtx.Dp(values.MarginPadding20)
+							gtx.Constraints.Min.X = gtx.Constraints.Max.X
+							return pg.materialLoader.Layout(gtx)
+						})
+					}),
+				)
```
itswisdomagain, line 556: We're already inside a flex layout, is it necessary to wrap the loader inside another flex layout with flex = 1??

ui/page/dcrdex/dex_onboarding_page.go
```diff
@@ -716,16 +764,12 @@ func (pg *DEXOnboarding) stepWaitForBondConfirmation(gtx C) D {
 					)
 				}),
 				layout.Rigid(func(gtx C) D {
-					return layout.Inset{Top: 10, Bottom: u10}.Layout(gtx, pg.Theme.Body1(values.StringF(values.StrDEXBondConfirmationMsg, "dex.decred.org", 2 /* TODO: use real values */)).Layout)
+					return layout.Inset{Top: 10, Bottom: u10}.Layout(gtx, pg.Theme.Body1(values.StringF(values.StrDEXBondConfirmationMsg, pg.selectedServerURL, pg.requiredBondConf)).Layout)
```
itswisdomagain, line 767: Is it possible for each step to have a state struct that holds all the variables for that step? I think that would be easier to follow than `pg` having many variables for different steps.

ui/page/dcrdex/dex_onboarding_page.go
itswisdomagain: Tested and it appears it is sufficient to just use layout.Center alone, without wrapping in a flex.

ui/page/dcrdex/dex_onboarding_page.go
```diff
@@ -880,76 +918,189 @@ func (pg *DEXOnboarding) HandleUserInteractions() {
...
+				pg.bondStrengthEditor.Editor.SetText(fmt.Sprintf("%d", minimumBondStrength))
+				pg.newTier = minimumBondStrength
 			}
 
-			// TODO: Validate server is reachable and connect.
-			_ = serverURL
-			_ = serverCert
+			if pg.isLoading {
+				break
+			}
```
itswisdomagain, line 1000: This can be before the `nextStep` var above?

ui/page/dcrdex/dex_onboarding_page.go
```diff
@@ -880,76 +918,189 @@ func (pg *DEXOnboarding) HandleUserInteractions() {
...
-			pg.currentStep = onboardingPostBond
-			pg.bondSourceWalletSelector = components.NewWalletAndAccountSelector(pg.Load /*, supportedAssets...  TODO: Use assets provided by selected DEX server. */).
-				Title(values.String(values.StrSelectWallet)).
-				WalletSelected(func(wm *load.WalletMapping) {
-					if err := pg.bondSourceAccountSelector.SelectFirstValidAccount(wm); err != nil {
-						log.Error(err)
+			dexInitialized := pg.dexc.IsInitialized()
+			if !dexInitialized {
+				go func() {
+					pg.isLoading = true
```
itswisdomagain, line 1005: Set before the goroutine.

ui/page/dcrdex/dex_onboarding_page.go
```diff
@@ -880,76 +918,189 @@ func (pg *DEXOnboarding) HandleUserInteractions() {
...
+					pg.dexPass = []byte(pg.passwordEditor.Editor.Text())
```
itswisdomagain, line 1006: Why do we need to cache the pass? Do we zero the bytes when we no longer need it?

ui/page/dcrdex/dex_onboarding_page.go
```diff
@@ -880,76 +918,189 @@ func (pg *DEXOnboarding) HandleUserInteractions() {
...
+			}
 
-			pg.currentStep = onboardingPostBond
-			pg.bondSourceWalletSelector = components.NewWalletAndAccountSelector(pg.Load /*, supportedAssets...  TODO: Use assets provided by selected DEX server. */).
-				Title(values.String(values.StrSelectWallet)).
-				WalletSelected(func(wm *load.WalletMapping) {
-					if err := pg.bondSourceAccountSelector.SelectFirstValidAccount(wm); err != nil {
-						log.Error(err)
+			dexInitialized := pg.dexc.IsInitialized()
+			if !dexInitialized {
```
itswisdomagain, line 1003: Should this be checking the current step instead?
UPDATE: I see we're inside `case onboardingChooseServer, onBoardingStepAddServer`, so should it even be possible for dexc not to be initialized at this stage?

itswisdomagain: Break this method into smaller methods, preferably for each step.

itswisdomagain (changes requested):
Looks like progress. Another review round.

Also, this PR title needs some love. It's implementing some dex functionality (specify), not adding live data to dex onboarding, as though there's some data somewhere that is being read and populated into the UI.

dexcore/core.go
```diff
@@ -0,0 +1,127 @@
...
+	stamp time.Time
+}
+
+// BondsFeeBuffer is a caching helper for the bonds fee buffer because
+// (*Core).BondsFeeBuffer returns a fresh fee buffer based on a current (but
+// padded) fee rate estimate. Values for a given asset are cached for 45
+// minutes. These values are meant to provide a sensible but well-padded fee
+// buffer for bond transactions now and well into the future, so a long expiry
+// is appropriate.
+func (dc *dexCore) BondsFeeBuffer(assetID uint32) (feeBuffer uint64) {
```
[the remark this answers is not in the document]
ukane-philemon: Okay.

ui/page/dcrdex/dex_onboarding_page.go
ukane-philemon: > Why do we need to cache the pass?

When users provide their pass via the pass modal, we shouldn't request the pass again while the user is on the onboarding page so we need to cache it so we can access it after the pass modal is closed.

ui/page/dcrdex/dex_onboarding_page.go
ukane-philemon: >  so should it even be possible for dexc not to be initialized at this stage?

Yes, dex is not initialized until the user confirms the server choice (i.e. if it's their first time).

ui/page/dcrdex/dex_onboarding_page.go
ukane-philemon: `core.DiscoverAccount` requires that dex has been initialised (i.e. pass has been set).

ui/page/components/bottom_nav.go
```diff
@@ -52,7 +52,7 @@ func (bottomNavigationbar *BottomNavigationBar) LayoutBottomNavigationBar(gtx la
 				layout.Rigid(func(gtx C) D {
 					list := layout.List{Axis: layout.Horizontal}
 					return list.Layout(gtx, len(bottomNavigationbar.BottomNaigationItems), func(gtx C, i int) D {
-						appWidth := gtx.Dp(bottomNavigationbar.Load.CurrentAppWidth())
+						appWidth := gtx.Dp(unit.Dp(bottomNavigationbar.Load.CurrentAppWidth))
```
JustinBeBoy, line 55: in load file you noted the CurrentAppWidth not inited, so why you change it from CurrentAppWidth() to CurrentAppWidth?

ui/page/root/wallet_selector_page.go
```diff
@@ -251,7 +252,7 @@ func (pg *WalletSelectorPage) pageContentLayout(gtx C) D {
 	}.Layout2(gtx, func(gtx C) D {
 		width := values.MarginPadding550
 		if pg.Load.IsMobileView() {
-			width = pg.Load.CurrentAppWidth()
+			width = unit.Dp(pg.Load.CurrentAppWidth)
```
JustinBeBoy, line 255: is the same above

JustinBeBoy (changes requested):

libwallet/assets_manager.go
```diff
@@ -780,3 +779,12 @@ func (mgr *AssetsManager) BlockExplorerURLForTx(assetType utils.AssetType, txHas
 func (mgr *AssetsManager) LogFile() string {
 	return filepath.Join(mgr.params.LogDir, LogFilename)
 }
+
+func (mgr *AssetsManager) IsAssetTypeSupported(assetType utils.AssetType) bool {
```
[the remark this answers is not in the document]
ukane-philemon: Yes, we do. I'll add a comment but this is a convenience method we use to identify assets (received from a dex server) that are not supported by crypto power.

ui/page/components/bottom_nav.go
ukane-philemon: Dang, this and the change below came from rebasing this PR. I'll take a closer look.

dreacot:
Screenshot from 2023-12-05 19-56-36

after i click start trading, i'm taken to the dex main page

JustinBeBoy (dismissed):
1. something happen when i click to recent order 

https://github.com/crypto-power/cryptopower/assets/19331824/48caee0d-3dc7-4de9-8ba6-afa4a86073ce

2.  Nothing happen when clicked to select dex server
bug1

3. validate input when add dex server
bug2

ui/page/dcrdex/dcrdex_page.go
```diff
@@ -85,17 +92,50 @@ func (pg *DEXPage) Layout(gtx C) D {
 			return pg.splashPage(gtx)
 		})
 	}
+	hasMultipleWallets := pg.isMultipleAssetTypeWalletAvailable()
+	privacyModeOff := pg.AssetsManager.IsHTTPAPIPrivacyModeOff(utils.ExchangeHTTPAPI)
```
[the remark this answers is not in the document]
ukane-philemon: Yes, both pages are purely informational and as such, there's no gain in having an overlay. The user would only need to see the overlay IF the want to trade by clicking on the btn. Then they can proceed to the settings from there.

ukane-philemon:
https://github.com/crypto-power/cryptopower/pull/238/commits/00de2a00c1bd898aa729f832648dafa5eeab43f4 is a lil out of scope for this PR but maybe we can have it here @dreacot?

JustinBeBoy:
> I’m having issues viewing the video and images atm but will get to it as soon as I can. Thanks for the reviews. RE 2: if there’s only one server what do you expect the dropdown should do when clicked? In other words, the only server listed is selected by default and nothing should be done when it's clicked.
> something happen when i click to recent order
> @JustinBeBoy, please create an issue cuz it's out of scope for this PR and I don have swaps to investigate with.

okay

JustinBeBoy:
> RE 2: if there’s only one server what do you expect the dropdown should do
when clicked?  In other words, the only server listed is selected by default and nothing should be done when it's clicked.

With normal user behavior, if you let them click, there must be some action, so I think it should open a dropdown even though there is only 1 option.

ukane-philemon:
> With normal user behavior, if you let them click, there must be some action, so I think it should open a dropdown even though there is only 1 option.

> @dreacot on matrix: No, most dropdowns should have an item selected by default

I think it's pointless opening a drop-down for one item.

dreacot:
i get this after clicking next on the select server step

Screenshot from 2023-12-07 12-37-19

dreacot:
i get this after clicking next on the select server step

Screenshot from 2023-12-07 12-37-19

JustinBeBoy:
> > i get this after clicking next on the select server step
> > Screenshot from 2023-12-07 12-37-19
> 
> Uhmm, I test server has been offline for some time. I'll see if I can replace it with another test server.

please do it, I can test the next step

ukane-philemon:
Note to reviewers: You'll only be able to post bond with #269.

ui/page/dcrdex/dex_onboarding_page.go
```diff
@@ -844,76 +892,207 @@ func (pg *DEXOnboarding) HandleUserInteractions() {
 				pg.currentStep = onboardingChooseServer
 			}
 		case onboardingChooseServer, onBoardingStepAddServer:
-			var serverURL string
-			var serverCert []byte
+			if pg.isLoading {
+				break
+			}
```
itswisdomagain, line 897: I think check this early and return if some background process is taking place, instead of checking in the switch cases. The first case doesn't check but that should also never run if some background process is taking place.

ui/page/dcrdex/dex_onboarding_page.go
```diff
@@ -844,76 +892,207 @@ func (pg *DEXOnboarding) HandleUserInteractions() {
...
+				FeeBuffer: pg.core.BondsFeeBuffer(bondAsset.ID),
 			}
 
-			// TODO: Post bond, wait for confirmations and redirect to market page.
-			pg.currentStep = onBoardingStepWaitForConfirmation
-			// Scroll to the top of the confirmation page after leaving the long
-			// post bond form.
-			pg.scrollContainer.Position.Offset = 0
+			pg.isLoading = true
+			pg.bondConfirmationInfo = new(bondConfirmationInfo)
```
itswisdomagain, line 1055: This should probably be nil until after `core.PostBond` completes and returns a response.

ui/page/dcrdex/dex_onboarding_page.go
```diff
@@ -844,76 +892,207 @@ func (pg *DEXOnboarding) HandleUserInteractions() {
...
 	ac := pg.bondSourceAccountSelector.SelectedAccount()
-	if ac.Balance.Spendable.ToCoin() < float64(pg.newTier)*20 /* TODO: Use actual bond cost + reservations + fees */ {
-		pg.bondSourceAccountSelector.SetError(values.String(values.StrInsufficientFundsInAccount))
+	asset := pg.bondSourceWalletSelector.SelectedWallet()
+	bondAsset := pg.bondServer.bondAssets[asset.GetAssetType()]
+	bondsFeeBuffer := pg.core.BondsFeeBuffer(bondAsset.ID)
+	bondCost := uint64(pg.newTier)*bondAsset.Amt*bondOverlap + bondsFeeBuffer
+	bondAmt := asset.ToAmount(int64(bondCost))
+	if ac.Balance.Spendable.ToInt() < bondAmt.ToInt() {
+		pg.bondSourceAccountSelector.SetError(values.StringF(values.StrInsufficientBondAmount, bondAmt.String()))
```
itswisdomagain, line 1095: Core doesn't have a method to check this? I'd rather defer this to core, even if it means the user won't see an error message until core.PostBond errors for insufficient funds.

ui/page/exchange/create_order_page.go
```diff
@@ -128,6 +130,8 @@ func NewCreateOrderPage(l *load.Load) *CreateOrderPage {
 			Alignment: layout.Middle,
 			Axis:      layout.Vertical,
 		}},
+		startTradingBtn: l.Theme.Button(values.String(values.StrStartTrading)),
+		isFirstVisit:    true,
```
itswisdomagain, line 134: Hmm, so always true until you click start trading?

ui/values/strings.go
```diff
@@ -845,6 +845,7 @@ const (
 	StrWalletStatus                    = "walletStatus"
 	StrWalletSyncing                   = "walletSyncing"
 	StrWalletToPurchaseFrom            = "walletToPurchaseFrom"
+	StrWarningVote                     = "warningVote"
```
itswisdomagain, line 848: Probably re-introduced in a rebase, no longer used.

ui/values/localizable/en.go
```diff
@@ -854,5 +854,9 @@ const EN = `
 "booked" = "Booked"
 "executed" = "Executed"
 "cannotSpendWatchOnlyWallet" = "You cannot spend from a watch only wallet, try creating another wallet."
-
+"multipleAssetTypeRequiredToTradeDEX" = "Multiple coin type wallets are required to trade on DCRDEX."
+"noSupportedBondAsset" = "DEX server (%s) does not support available asset wallets. Please choose another server."
+"dexPassword" = "DCRDEX Password"
+"insufficientBondAmount" = "Insufficient funds in this account, you a minimum of %v to post bond."
```
itswisdomagain, line 860:
```suggestion
"insufficientBondAmount" = "Insufficient funds in this account, you need a minimum of %v to post bond."
```

ui/preference/list_preference.go
```diff
@@ -111,8 +111,9 @@ func (lp *ListPreferenceModal) SavePreferenceKeyedValue() {
 	case sharedW.CurrencyConversionConfigKey:
 		lp.AssetsManager.SetCurrencyConversionExchange(val)
 	case sharedW.LanguagePreferenceKey:
+		// TODO: We should be able to update dex core's language when the user
+		// changes language.
 		lp.AssetsManager.SetLanguagePreference(val)
-	case sharedW.LogLevelConfigKey:
```
itswisdomagain, line 115: Undo?

itswisdomagain (dismissed):
If you had an idea to put `dexc` in AssetsManager, let's do that. I think it can work now.

ui/preference/list_preference.go
ukane-philemon: Dang. Yh.

ui/page/exchange/create_order_page.go
ukane-philemon: Yes. We don't; save to asset manager db atm.

dreacot (approved):

ukane-philemon:
> I'm wondering if the back button can't go back, should it be here, and from step 2 can we go back to step 1?

Password is set the first time we try to onboard. Users that abandon the process can't set pass twice.

ui/page/dcrdex/dex_onboarding_page.go
ukane-philemon: > Core doesn't have a method to check this?

Core will just fail when it attempts to make payments but I don't think it's a good UX for users to fail before they know they can choose another wallet with sufficient funds.

dreacot:
> > I'm wondering if the back button can't go back, should it be here, and from step 2 can we go back to step 1?
> 
> Password is set the first time we try to onboard. Users that abandon the process can't set pass twice.

i think he is saying th eback button shouldn't be there since users can't go back to step  1

ukane-philemon:
> i think he is saying th eback button shouldn't be there since users can't go back to step 1

okay, what do you think?

dreacot:
> > i think he is saying th eback button shouldn't be there since users can't go back to step 1
> 
> okay, what do you think?

let's leave it for now

dreacot (approved):

PR #250, opened 2023-11-13
Multi: Refactor MasterPage and PageStack

### LINKED ISSUE #249: app.NewMasterPage should accept a start page
See: https://github.com/crypto-power/cryptopower/pull/240#discussion_r1387199643

ukane-philemon:
Closes #249 

-- Refactor app.NewMasterPage to accept and handle a start page.
-- Refactor *app.MasterPage.Display()
-- Refactor app.PageStack.Push to only prepare the next page.
-- Refactor pages that embed app.Masterpage.
-- [Refactor app.PageStack.Pop()](https://github.com/crypto-power/cryptopower/commit/5cc572e5e721e009d6b3e49f70ceae32cf2a16fc)
-- Fix some TODOs.

### DIFF
--- a/app/pagestack.go
+++ b/app/pagestack.go
@@ -41,14 +41,6 @@ func (pageStack *PageStack) PushAndNavigate(newPage Page, navigator PageNavigato
 	pageStack.mtx.Lock()
 	defer pageStack.mtx.Unlock()
 
-	if l := len(pageStack.pages); l > 0 {
-		currentPage := pageStack.pages[l-1]
-		if currentPage.ID() == newPage.ID() {
-			return false
-		}
-		currentPage.OnNavigatedFrom()
-	}
-
 	// Close any previous instance of this type. Use the Closed() method if
 	// implemented, to signal that the instance will never be re-displayed.
 	for i, existingPage := range pageStack.pages {
@@ -63,8 +55,6 @@ func (pageStack *PageStack) PushAndNavigate(newPage Page, navigator PageNavigato
 	}
 
 	pageStack.pages = append(pageStack.pages, newPage)
-	newPage.OnAttachedToNavigator(navigator)
-	newPage.OnNavigatedTo()
 	return true
 }
 
--- a/app/pagestack.go
+++ b/app/pagestack.go
@@ -48,6 +48,14 @@ func (pageStack *PageStack) PushAndNavigate(newPage Page, navigator PageNavigato
 	pageStack.mtx.Lock()
 	defer pageStack.mtx.Unlock()
 
+	if l := len(pageStack.pages); l > 0 {
+		currentPage := pageStack.pages[l-1]
+		if currentPage.ID() == newPage.ID() {
+			return false
+		}
+		currentPage.OnNavigatedFrom()
+	}
+
 	// Close any previous instance of this type. Use the Closed() method if
 	// implemented, to signal that the instance will never be re-displayed.
 	for i, existingPage := range pageStack.pages {
@@ -62,6 +70,8 @@ func (pageStack *PageStack) PushAndNavigate(newPage Page, navigator PageNavigato
 	}
 
 	pageStack.pages = append(pageStack.pages, newPage)
+	newPage.OnAttachedToNavigator(navigator)
+	newPage.OnNavigatedTo()
 	return true
 }
 
--- a/app/pagestack.go
+++ b/app/pagestack.go
@@ -13,11 +13,18 @@ type PageStack struct {
 	pages []Page
 }
 
-// NewPageStack creates a new PageStack object.
-func NewPageStack(name string) *PageStack {
-	return &PageStack{
+// NewPageStack creates a new PageStack object. startPage is optional but should
+// have it's navigator already attached if provided.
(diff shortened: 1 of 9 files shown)


app/pagestack.go
```diff
@@ -32,23 +32,25 @@ func (pageStack *PageStack) Top() Page {
...
-// instances of the same page from the stack. An about-to-display signal is sent
-// to the new page via newPage.OnNavigatedTo() while page.OnNavigatedFrom() is
-// called on the current page to signal that the current page is no longer the
-// displayed page.
-func (pageStack *PageStack) Push(newPage Page, navigator PageNavigator) bool {
+// instances of the same page from the stack.
+func (pageStack *PageStack) Push(newPage Page) bool {
+	if currentPage := pageStack.Top(); currentPage != nil && currentPage.ID() == newPage.ID() {
+		return false
+	}
```
itswisdomagain, line 49: What's the motivation for the change to this method? This change introduces a potential issue. If the current page has the same ID as the new page, nothing is done by this method (which was the behaviour before this change), but the problem is, before this method was called, `NavigateFromCurrentPage()` was called. And so the current page would basically be useless now even though it remains on the top of the stack.

Discuss the motivation for the additional changes in this PR before making further changes otherwise revert to the previous code to avoid regressions.

itswisdomagain (changes requested):

app/pagestack.go
ukane-philemon: Dang. NavigateFromCurrentPage should be after the paste has been pushed and Push returned true.

I’m suggesting to reduce pageStack.Push responsibility because now that we have a start page when creating a master page, only callers can accurately tell when the new page is ready for display.

itswisdomagain: > only callers can accurately tell when the new page is ready for display

Don't understand this. Why is this so?

ukane-philemon: Now use Push in NewMasterPage and in Display. NewMasterPage cares only about preparing the page(I.e adding it to the page stack and binding the navigator to the start page) while Display does everything including notifying the page it is read for display by calling the OnNavigateTo method on a new page.

We will end up calling the top page OnNavigateTo method twice, If we continue to allow pageStack.Push to notify pages that they are ready (the first would be in pageStack.Push and the second would be when the MasterPage OnNavigatedTo Method is called).

Calling the OnNavigateTo method of a page twice in a roll (without calling OnNavigateFrom) is not without consequence, one of which is the fact page notifiers will be registered twice.

itswisdomagain: I'm still confused.

ukane-philemon: Typo

ukane-philemon: Please see how NewMasterPage is used in this PR and masterpage.Display too.

ui/page/root/trade_page.go
```diff
@@ -86,16 +82,15 @@ func (pg *TradePage) OnNavigatedTo() {
...
+		}
 	}
 
-	if pg.tab.SelectedIndex() == 0 {
-		pg.Display(dcrdex.NewDEXPage(pg.Load))
-	}
-	if pg.tab.SelectedIndex() == 1 {
-		pg.Display(exchange.NewCreateOrderPage(pg.Load))
-	}
+	pg.CurrentPage().HandleUserInteractions()
```
ukane-philemon, line 93: Forwarding User Interaction should happen `after` a new page has been assigned if we hit the tab changed if branch. Same thing applies to other master pages.

ukane-philemon:
> @ukane-philemon i'm not fully sure what the scope of this PR aims to achieve or it's necessity

Mainly this "Refactor app.NewMasterPage to accept and handle a start page.". This way we don't have to check `if pg.CurrentPage() == nil` everywhere when we can initialize with a page.

JustinBeBoy:
> > @ukane-philemon i'm not fully sure what the scope of this PR aims to achieve or it's necessity
> 
> Mainly this "Refactor app.NewMasterPage to accept and handle a start page.". This way we don't have to check `if pg.CurrentPage() == nil` everywhere when we can initialize with a page.

I agree with @dreacot , I'm not sure about the purpose of this PR, your PR and the current PR are not much different.

- Current code: Whenever we create a new master page, obviously `pg.CurrentPage()` will be nil, instead of checking we will always create a page for it.
- Your PR: Add a page when creating the master page but that is optional, it does not prevent `pg.CurrentPage()` from being nil if the page is not initialized for the master page, If the page passed in is mandatory then there would be no need to check `pg.CurrentPage() == nil`

This PR I see is just another approach but the problem is still not solved as you said

ukane-philemon:
> Your PR: Add a page when creating the master page but that is optional, it does not prevent pg.CurrentPage() from being nil if the page is not initialized for the master page, If the page passed in is mandatory then there would be no need to check pg.CurrentPage() == nil


Yes, most of the master pages don't need to check as they always have a start page. Only the Window's masterPage and dex masterPage check for a nill CurrentPage() and with #238, we can tell which page to display from the onset for the dex masterPage.

dreacot:
moving to draft

ukane-philemon:
I'll resubmit if this is needed in the future.

PR #267, opened 2023-11-25
Rename pageStack.Push to be more intuitive

### LINKED ISSUE #252: Refactor app.PageStack.Push(...)
Currently, `app.PageStack.Push(...)` does more than push a new page to the page stack, it also binds the page to a navigator and executes the `OnNavigateTo` method on the page. IMO, `Push` should do nothing more than its name suggests. Removing duplicates before adding a new page in pageStack.Push is acceptable but I think it's better and more intuitive for its caller to bind the page as required and execute the `OnNavigateTo` method.

Any thoughts @dreacot? I have some code suggestions we can discuss if we're okay with this refactor.

ukane-philemon:
Closes #252

### DIFF
--- a/app/pagestack.go
+++ b/app/pagestack.go
@@ -32,12 +32,11 @@ func (pageStack *PageStack) Top() Page {
 	return nil
 }
 
-// Push pushes the specified page to the top of the stack, removing all other
-// instances of the same page from the stack. An about-to-display signal is sent
-// to the new page via newPage.OnNavigatedTo() while page.OnNavigatedFrom() is
-// called on the current page to signal that the current page is no longer the
-// displayed page.
-func (pageStack *PageStack) Push(newPage Page, navigator PageNavigator) bool {
+// PushAndNavigate pushes the specified page to the top of the stack, removing
+// all other instances of the same page from the stack and executes the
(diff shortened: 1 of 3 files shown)


app/pagestack.go
```diff
@@ -32,12 +32,11 @@ func (pageStack *PageStack) Top() Page {
...
 
-// Push pushes the specified page to the top of the stack, removing all other
-// instances of the same page from the stack. An about-to-display signal is sent
-// to the new page via newPage.OnNavigatedTo() while page.OnNavigatedFrom() is
-// called on the current page to signal that the current page is no longer the
-// displayed page.
-func (pageStack *PageStack) Push(newPage Page, navigator PageNavigator) bool {
+// PushAndNavigate pushes the specified page to the top of the stack, removing
+// all other instances of the same page from the stack and executes the
+// OnNavigatedTo() on the newPage while page.OnNavigatedFrom() is called on the
```
itswisdomagain, line 37:
```suggestion
// OnNavigatedTo() method on the newPage while page.OnNavigatedFrom() is called on the
```

itswisdomagain (approved):

dreacot (approved):

PR #268, opened 2023-11-26
Improve infinite scroll list functionality

### LINKED ISSUE #93: the pagination setup on the transactions page is buggy
from the screenshot, it's clear that there are 103 mixed transactions, but after scrolling through the pages while, all is selected, then switching to sent transactions and switching to mixed transactions the list isn't populated

Screenshot 2023-09-25 at 1 54 08 PM

if i time my scrolling well enough that the scroll bar start at the top on the next page, i'm unable to return to the previous page

Screenshot 2023-09-25 at 1 56 21 PM

Sirmorrison:
Closes #147 
Fixes #93 
and a follow up for #198

This PR improve the current list scroll functionality. It allows data to be continuously updated, without an obvious need to clear the already existing data, up until a maximum list size is attained, after which the first Nth items on the list are trimmed off.

It fixes issues with list items not being reset until the next iteration when `FetchScrollData()` is called.

It Fixes issues with the app freezing when the last item on a scroll list is reached and a user tries to preform more downwards sccroll.

### DIFF
--- a/ui/page/components/items_scroll.go
+++ b/ui/page/components/items_scroll.go
@@ -60,8 +60,8 @@ func NewScroll[T any](load *load.Load, pageSize int32, queryFunc ScrollFunc[T]) 
 
 // FetchScrollData is a mutex protected fetchScrollData function. At the end of
 // the function call a window reload is triggered. Returns that latest records.
-// isReset is used to reset the offset value.
-func (s *Scroll[T]) FetchScrollData(isReset bool, window app.WindowNavigator) {
+// If reset is true, the offset value will be reset to 0.
+func (s *Scroll[T]) FetchScrollData(reset bool, window app.WindowNavigator) {
 	s.mu.Lock()
 	// s.data is not nil when moving from details page to list page.
 	if s.data != nil {
@@ -70,9 +70,8 @@ func (s *Scroll[T]) FetchScrollData(isReset bool, window app.WindowNavigator) {
 		s.offset -= s.pageSize
 	}
 	s.mu.Unlock()
-	// set isReverse to default false as callers of this method are not
-	// perform a reverse scroll action
-	s.fetchScrollData(false, isReset, window)
+	// set isReverse = false since this method is only called when scrolling down
+	s.fetchScrollData(false, reset, window)
 }
 
 // fetchScrollData fetchs the scroll data and manages data returned depending on
@@ -80,11 +79,10 @@ func (s *Scroll[T]) FetchScrollData(isReset bool, window app.WindowNavigator) {
 // the page, all the old data is replaced by the new fetched data making it
 // easier and smoother to scroll on the UI. At the end of the function call
 // a window reload is triggered.
-func (s *Scroll[T]) fetchScrollData(isReverse, isReset bool, window app.WindowNavigator) {
+func (s *Scroll[T]) fetchScrollData(isReverse, reset bool, window app.WindowNavigator) {
 	s.mu.Lock()
 
-	if isReset {
-		// resets the values for use on the next iteration.
+	if reset {
 		s.resetList()
 	}
 
@@ -100,7 +98,7 @@ func (s *Scroll[T]) fetchScrollData(isReverse, isReset bool, window app.WindowNa
 	if isReverse {
 		s.offset -= s.pageSize
 	} else {
-		if s.data != nil && !isReset {
+		if s.data != nil && !reset {
 			s.offset += s.pageSize
 		}
 	}
--- a/ui/page/components/items_scroll.go
+++ b/ui/page/components/items_scroll.go
@@ -13,8 +13,8 @@ import (
 )
 
 // ScrollFunc is a query function that accepts offset and pagesize parameters and
-// returns data interface, count of the items in the data interface and an error.
-type ScrollFunc[T any] func(offset, pageSize int32) (data []T, count int, err error)
+// returns data interface and an error.
+type ScrollFunc[T any] func(offset, pageSize int32) (data []T, err error)
 
 type Scroll[T any] struct {
 	load      *load.Load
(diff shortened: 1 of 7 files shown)


ui/page/components/items_scroll.go
```diff
@@ -67,47 +70,48 @@ func (s *Scroll[T]) FetchScrollData(isReverse bool, window app.WindowNavigator)
 		s.offset -= s.pageSize
 	}
 	s.mu.Unlock()
-
-	s.fetchScrollData(isReverse, window)
+	// set isReverse to default false as callers of this method are not
+	// perform a reverse scroll action
```
itswisdomagain, line 74:
```suggestion
	// set isReverse = false since this method is only called when scrolling down
```

ui/page/components/items_scroll.go
```diff
@@ -67,47 +70,48 @@ func (s *Scroll[T]) FetchScrollData(isReverse bool, window app.WindowNavigator)
...
+	s.fetchScrollData(false, isReset, window)
 }
 
 // fetchScrollData fetchs the scroll data and manages data returned depending on
 // on the up or downwards scrollbar movement. Unless the data fetched is less than
 // the page, all the old data is replaced by the new fetched data making it
 // easier and smoother to scroll on the UI. At the end of the function call
 // a window reload is triggered.
-func (s *Scroll[T]) fetchScrollData(isReverse bool, window app.WindowNavigator) {
+func (s *Scroll[T]) fetchScrollData(isReverse, isReset bool, window app.WindowNavigator) {
```
itswisdomagain, line 83:
```suggestion
func (s *Scroll[T]) fetchScrollData(isReverse, reset bool, window app.WindowNavigator) {
```

ui/page/components/items_scroll.go
```diff
@@ -58,7 +60,8 @@ func NewScroll[T any](load *load.Load, pageSize int32, queryFunc ScrollFunc[T])
 
 // FetchScrollData is a mutex protected fetchScrollData function. At the end of
 // the function call a window reload is triggered. Returns that latest records.
-func (s *Scroll[T]) FetchScrollData(isReverse bool, window app.WindowNavigator) {
+// isReset is used to reset the offset value.
+func (s *Scroll[T]) FetchScrollData(isReset bool, window app.WindowNavigator) {
```
itswisdomagain, line 64:
```suggestion
// If reset is true, the offset value will be reset to 0.
func (s *Scroll[T]) FetchScrollData(reset bool, window app.WindowNavigator) {
```

ui/page/components/items_scroll.go
```diff
@@ -67,47 +70,48 @@ func (s *Scroll[T]) FetchScrollData(isReverse bool, window app.WindowNavigator)
...
 // the page, all the old data is replaced by the new fetched data making it
 // easier and smoother to scroll on the UI. At the end of the function call
 // a window reload is triggered.
-func (s *Scroll[T]) fetchScrollData(isReverse bool, window app.WindowNavigator) {
+func (s *Scroll[T]) fetchScrollData(isReverse, isReset bool, window app.WindowNavigator) {
 	s.mu.Lock()
 
+	if isReset {
+		// resets the values for use on the next iteration.
+		s.resetList()
```
itswisdomagain, line 88: Are the reset values used on the next iteration or right away?

ui/page/components/items_scroll.go
```diff
@@ -67,47 +70,48 @@ func (s *Scroll[T]) FetchScrollData(isReverse bool, window app.WindowNavigator)
...
+	}
+
+	// reset loaded items if we have loaded all items and we are scrolling back to the top
+	if isReverse && s.loadedAllItems {
+		s.loadedAllItems = false
+	}
```
itswisdomagain, line 94: Why? If all items were already loaded, scrolling back up doesn't delete the items that were already loaded, does it?

itswisdomagain (changes requested):
I really don't yet understand what problem this PR is meant to solve, so stopping my review here.

ui/page/components/items_scroll.go
Sirmorrison: Right away. Reason is, most callers of the method needing a reset, either performed a filter change or completed sync.

ui/page/components/items_scroll.go
```diff
@@ -58,7 +60,8 @@ func NewScroll[T any](load *load.Load, pageSize int32, queryFunc ScrollFunc[T])
 
 // FetchScrollData is a mutex protected fetchScrollData function. At the end of
 // the function call a window reload is triggered. Returns that latest records.
-func (s *Scroll[T]) FetchScrollData(isReverse bool, window app.WindowNavigator) {
+// If reset is true, the offset value will be reset to 0.
+func (s *Scroll[T]) FetchScrollData(reset bool, window app.WindowNavigator) {
```
itswisdomagain, line 64: Improve this comment. This is an exported method but it doesn't say what it does. Instead, it references a private method. Please update this comment with a clear description of what this method does.

ui/page/components/items_scroll.go
```diff
@@ -58,7 +60,8 @@ func NewScroll[T any](load *load.Load, pageSize int32, queryFunc ScrollFunc[T])
...
 	s.mu.Lock()
 	// s.data is not nil when moving from details page to list page.
 	if s.data != nil {
```
itswisdomagain, line 67: What does this comment mean? What's happening inside this if statement?
It's unclear code like these that leads to bugs.

If you fully understand all the code in this file, please add generous comments in as many places as possible.
If you don't fully understand what's happening here, then you shouldn't be maintaining or improving it. We should probably be talking about a new struct that achieves the intended objectives, but is much clearer and easier to read and understand.

itswisdomagain (changes requested):
I'm going to be honest and say most of the code in `ui/page/components/items_scroll.go` don't make sense to me.

How can we make it make sense? If I have to spend a lot of time to understand a few lines of code, that's a waste of time.

Issue #297, opened 2023-11-30
the mobile send and receive buttons are showing on desktop


dreacot:
![Screenshot from 2023-11-29 18-25-57](https://github.com/crypto-power/cryptopower/assets/25265396/24b5a251-a6fe-436a-b930-5f8fcefde11a)

PR #298, opened 2023-11-30
fix mobile buttons showing on desktop

dreacot:
closes https://github.com/crypto-power/cryptopower/issues/297

before:
Screenshot from 2023-11-29 18-25-57

this PR

Screenshot from 2023-11-29 18-28-01

### DIFF
--- a/ui/page/root/home_page.go
+++ b/ui/page/root/home_page.go
@@ -522,9 +522,10 @@ func (hp *HomePage) totalBalanceLayout(gtx C) D {
 			}),
 			layout.Rigid(hp.balanceLayout),
 			layout.Rigid(func(gtx C) D {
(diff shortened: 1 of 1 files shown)


ui/page/root/home_page.go
```diff
@@ -523,90 +523,90 @@ func (hp *HomePage) totalBalanceLayout(gtx C) D {
 			layout.Rigid(hp.balanceLayout),
 			layout.Rigid(func(gtx C) D {
 				if hp.Load.IsMobileView() {
```
itswisdomagain, line 525: I think the diff will be easier to follow if you reverse the if statement and return early.
```suggestion
				if !hp.Load.IsMobileView() {
```

PR #299, opened 2023-11-30
add mobile build instructions

### LINKED ISSUE #111: README: Add mobile build instructions.
The readme could use build instructions for android and ios.

dreacot:
closes #111

### DIFF
--- a/how-to-build-mobile.md
+++ b/how-to-build-mobile.md
@@ -4,7 +4,13 @@ This readme assumes you have a working Android or iOS environment.
 
 ## 1. Building for Android
 
-Note: You need to have [gogio](https://gioui.org/doc/install/android) to build.
+Note: To build Cryptopower for Android, you need to have;
+
+1.  gogio
+2.  [Android SDK with NDK bundle](https://developer.android.com/tools)
+3.  The ANDROID_SDK_ROOT point to the SDK root directory e.g `export ANDROID_SDK_ROOT=$HOME/.local/share/Android/Sdk`
+
+Proceed to the [gioui Android doc](https://gioui.org/doc/install/android) for more info regarding gogio and other dependencies you will be needing.
 
 execute the command below in a terminal window to install gogio:
 
@@ -16,13 +22,18 @@ cd to the cryptopower root directory and execute the command below to generate a
 
 there should now be a cryptopower.apk file in the cryptopower root directory. You can send this file to your android device and install it.
 
-or to send it to your device autmatically, execute the command below:
+or to send it to your device automatically, execute the command below:
 
 `adb install cryptopower.apk`
 
 ## 2. Building for iOS
 
-Note: You need to have [gogio](https://gioui.org/doc/install/ios) to build.
+Note: To build Cryptopower for iOS, you need to have;
+
+1. gogio
+2. [Xcode](https://developer.apple.com/xcode/)
+
+Proceed to the [gioui iOS doc](https://gioui.org/doc/install/ios) for more info regarding gogio and other dependencies you will be needing.
 
 execute the command below in a terminal window to install gogio:
 
@@ -34,6 +45,6 @@ cd to the cryptopower root directory and execute the command below to generate a
 
 there should now be a cryptopower.app file in the cryptopower root directory. You can send this file to your iOS simulator.
 
-or to send it to your simulator autmatically, execute the command below:
+or to send it to your simulator automatically, execute the command below:
 
 `xcrun simctl install booted cryptopower.app`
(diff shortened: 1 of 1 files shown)


how-to-build-mobile.md
```diff
@@ -0,0 +1,39 @@
...
+
+Note: You need to have [gogio](https://gioui.org/doc/install/android) to build.
+
+execute the command below in a terminal window to install gogio:
+
+`go install gioui.org/cmd/gogio@latest`
+
+cd to the cryptopower root directory and execute the command below to generate a .apk file:
+
+`gogio -target android .`
```
JoeGruffins, line 15: I need to do `export ANDROID_SDK_ROOT=$HOME/.local/share/Android/Sdk` which might be useful to mention. Also could mention building with the `--testnet` flag for testnet

JoeGruffins (approved):
Could add what versions you need. For example, I was not able to build at first because my gcc and androind ndk were too high versioned. I need to use the ndk version 25.2.9519653 It tries to use a higher version if available, so I have been moving the higher version in the folder at `$HOME/.local/share/Android/Sdk` which probably isnt the correct way but I am unsure of how to do it otherwise. I also had to downgrade my gcc, it needs to be version 10. This too I downgrade by moving around files, I place the version 10 `g++` and `gcc` in `/usr/local/bin` which overrides the installed versions. I guess this info is helpful, it's what I have to do currently.

how-to-build-mobile.md
```diff
@@ -0,0 +1,39 @@
+# Building Cryptopower Wallet for Mobile
+
+This readme assumes you have a working Android or iOS environment.
+
+## 1. Building for Android
+
+Note: You need to have [gogio](https://gioui.org/doc/install/android) to build.
```
itswisdomagain, line 7: The link points to the android build instructions page which begins with a mention of android dependencies. The link text should probably say `[some dependencies, including gogio]` rather than just `[gogio]`.

how-to-build-mobile.md
itswisdomagain: I also get a prompt to `set ANDROID_SDK_ROOT to the Android SDK path` when I run this command. I have the path set in `$ANDROID_HOME` so I just do `ANDROID_SDK_ROOT=$ANDROID_HOME gogio -target android .`

itswisdomagain (approved):
Pretty straightforward. Just some nits.

Also, can you change `var isAndroidOS` to `const isAndroidOS` in `os.go` and `os_android.go` files?

how-to-build-mobile.md
dreacot: > Also could mention building with the --testnet flag for testnet

good catch

JoeGruffins: I do not have it set but Android Studio runs without it set. There is a clear message when it fails if it is not set though, so it's fine to not mention here I think if you do not want to.

itswisdomagain: Yeah, same here. My setup works without the `ANDROID_SDK_ROOT` env variable. I think that variable name is now `ANDROID_HOME` but gogio still looks for `ANDROID_SDK_ROOT`. Can't remember, but saw something similar somewhere sometime while trying to build with gogio the first time.

how-to-build-mobile.md
ukane-philemon: @dreacot.

how-to-build-mobile.md
```diff
@@ -0,0 +1,39 @@
...
+
+`go install gioui.org/cmd/gogio@latest`
+
+cd to the cryptopower root directory and execute the command below to generate a .app file:
+
+`gogio -o cryptopower.app -target ios .`
+
+there should now be a cryptopower.app file in the cryptopower root directory. You can send this file to your iOS simulator.
+
+or to send it to your simulator autmatically, execute the command below:
```
ukane-philemon, line 37: nit: `autmatically` -> `automatically`

how-to-build-mobile.md
```diff
@@ -0,0 +1,39 @@
...
+
+`go install gioui.org/cmd/gogio@latest`
+
+cd to the cryptopower root directory and execute the command below to generate a .apk file:
+
+`gogio -target android .`
+
+there should now be a cryptopower.apk file in the cryptopower root directory. You can send this file to your android device and install it.
+
+or to send it to your device autmatically, execute the command below:
```
ukane-philemon, line 19: nit: `autmatically` -> `automatically`

ukane-philemon (changes requested):

ukane-philemon (approved):
LGTM.

ukane-philemon (approved):

PR #200, opened 2023-10-24
Fix crash when import watch-only wallet and click send button

JustinBeBoy:
Resolve #182 

This PR fix:
- App crash when import only watch-only wallet.
- App crash when click send button
- App crash when import mixed account of watch-only wallet
- App crash when delete wallet.

pr1

PR #302, opened 2023-11-30
Hide send tab option if wallet is watch only

Sirmorrison:
Fix #200 

This PR hides the send and receive tab options when the selected wallet is a watch only wallet.


dreacot:
A watch only wallet is able to receive funds

PR #336, opened 2023-12-16
Implement mobile account page

### LINKED ISSUE #276: [Mobile]: add accounts page
use figma!

Image

ukane-philemon:
Closes #276

I think this UI is good enough and should be used on desktops too. Any thoughts?

This PR:
Screenshot 2023-12-16 at 3 09 31 AM

Figma:
Screenshot 2023-12-16 at 3 09 56 AM

### DIFF
--- a/ui/cryptomaterial/clickable.go
+++ b/ui/cryptomaterial/clickable.go
@@ -63,7 +63,7 @@ func (cl *Clickable) Enabled() bool {
 // LayoutWithInset draws a layout of a clickable and applies an hover effect if
 // an hover action is detected. rightInset and bottomInset are used to restrict
 // hover layout and should be supplied ONLY if a right or bottom inset/margin
-// was applied to w. 
+// was applied to w.
 func (cl *Clickable) LayoutWithInset(gtx C, w layout.Widget, rightInset, bottomInset unit.Dp) D {
 	return cl.button.Layout(gtx, func(gtx layout.Context) layout.Dimensions {
 		return layout.Stack{}.Layout(gtx,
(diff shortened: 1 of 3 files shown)


ui/cryptomaterial/clickable.go
```diff
@@ -60,19 +60,28 @@ func (cl *Clickable) Enabled() bool {
...
+// an hover action is detected. rightInset and bottomInset are used to restrict
+// hover layout and should be supplied ONLY if a right or bottom inset/margin
+// was applied to w. 
+func (cl *Clickable) LayoutWithInset(gtx C, w layout.Widget, rightInset, bottomInset unit.Dp) D {
 	return cl.button.Layout(gtx, func(gtx layout.Context) layout.Dimensions {
 		return layout.Stack{}.Layout(gtx,
 			layout.Expanded(func(gtx layout.Context) layout.Dimensions {
+				// Only hover on a widget, ignore inset or margin applied.
+				gtx.Constraints.Min.Y = gtx.Constraints.Min.Y - gtx.Dp(bottomInset)
+				gtx.Constraints.Min.X = gtx.Constraints.Min.X - gtx.Dp(rightInset)
```
ukane-philemon: @dreacot, I need to also support accounting for top and left inset so hover effects always stay within expected constraints. Any thoughts?

dreacot:
> I think this UI is good enough and should be used on desktops too. Any thoughts? 

sure we can use this on desktop too

ui/cryptomaterial/clickable.go
dreacot: seems to work okay

dreacot (approved):

PR #345, opened 2023-12-20
add binary build workflow

dreacot:


.github/workflows/build.yml
```diff
@@ -0,0 +1,69 @@
+name: Cross-Platform Build on New Tag
+
+on:
+  pull_request:
+  push:
+    tags:
+      - '*'
```
itswisdomagain, line 7: Seems it runs on PRs too, not just tags.

dreacot: Yeah, i added the pr trigger there, so I wouldn't have to be creating and deleting tags, plus so i can test without merging, once i get all binaries building successfully i’d remove the pr trigger

PR #369, opened 2024-01-08
Sync: Release write mutex on sync data.

### LINKED ISSUE #357: [Mobile] App freezes when turn on sync the wallet
Some mobile devices freeze when turn on sync the wallet. Need to optimize the sync for mobile mode

JoeGruffins:
May close #357 @JustinBeBoy can you try this out

### DIFF
--- a/libwallet/assets/btc/sync.go
+++ b/libwallet/assets/btc/sync.go
@@ -152,7 +152,6 @@ func (asset *Asset) bestServerPeerBlockHeight() {
 
 func (asset *Asset) updateSyncProgress(rawBlockHeight int32) {
 	asset.syncData.mu.Lock()
-	defer asset.syncData.mu.Unlock()
 
 	// Update the best block synced in the connected peers if need be
 	asset.bestServerPeerBlockHeight()
@@ -164,6 +163,7 @@ func (asset *Asset) updateSyncProgress(rawBlockHeight int32) {
 		asset.syncData.headersFetchProgress.StartHeaderHeight = &rawBlockHeight
 
 		if asset.syncData.bestBlockheight != rawBlockHeight {
+			asset.syncData.mu.Unlock()
 			// A rescan progress update must have been sent. Allow it
 			return
 		}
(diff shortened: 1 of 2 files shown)


libwallet/assets/btc/sync.go
```diff
@@ -191,26 +191,30 @@ func (asset *Asset) updateSyncProgress(rawBlockHeight int32) {
...
+	asset.syncData.mu.Unlock()
 
 	// publish the sync progress results to all listeners.
+	asset.syncData.mu.RLock()
 	for _, listener := range asset.syncData.syncProgressListeners {
 		if listener.OnHeadersFetchProgress != nil {
 			listener.OnHeadersFetchProgress(&asset.syncData.headersFetchProgress)
 		}
 	}
+	asset.syncData.mu.RUnlock()
```
JoeGruffins, line 203: This is the problem area causing the freezes. updateSyncProgress in holds the syncData lock until its finished. At then end it does listener.OnHeadersFetchProgress which the info page does a updateSyncProgress which does pg.fetchSyncProgress which does pg.deleteSyncProgress which tries wal.IsSynced which needs a syncData read lock that it cannot get.

libwallet/assets/btc/sync.go
```diff
@@ -191,26 +191,30 @@ func (asset *Asset) updateSyncProgress(rawBlockHeight int32) {
...
 	asset.syncData.syncing = false
+	asset.syncData.mu.Unlock()
+
+	asset.handleSyncUIUpdate()
 }
 
 func (asset *Asset) handleSyncUIUpdate() {
+	asset.syncData.mu.RLock()
+	defer asset.syncData.mu.RUnlock()
 	for _, listener := range asset.syncData.syncProgressListeners {
```
JoeGruffins, line 218: I don't think this was causing any problems but it follows the same problem pattern causing the freezes so changing just in case. If a read lock is required in OnSyncCompleted that is now OK.

JustinBeBoy:
It looks good.
I'm not sure if this PR will prevent the App from freezing, there aren't any logs recorded so it's impossible to determine the cause, but nonetheless the changes in the PR should increase application performance.

good job @JoeGruffins

JustinBeBoy (approved):

Issue #367, opened 2024-01-05
observations


dreacot:
1. Screens are greyed out with msg that says 'until sync is complete.' but sync isnt even started. Should probably say that the wallet sync needs to be started or something to that regards. Like, click here to start syncing. and while syncing the msg could say something about the current sync progress %
2. Tab key on 'Enter wallet details' page doesnt work as expected. Also, should be Enter wallet passphrase since thats what the page is....not details.
3. Wallets page list doesnt have consistent use of decimal places.
4. Delay whenever clicking on Overview tab and then trying to click to another tab. Doesnt let me switch to any other tabs for about 4-5 seconds.
5. Settings button has delay before loading settings view.
6. Send to wallet account selector is broken, shows no accounts in list.
7. Annoying left to right scroll buttons on top even though there is unused width space on my app window.
8. Also...why are the buttons extra wide with white space if they are just going to be hidden by scroll buttons?
9. I already had a wallet created, but this doesnt see it. Does this mean migration is broken from 1.0?
10. Recommended and Advanced options use the same outline color for selected and hover. Hard to tell whats selected.
11. Also no idea what Im actually deciding by doing Recommended or Advanced, and no idea what happened by selecting it. Either way it just goes to the main page.
12. Send, Receive and Exchange Decred, Bitcoin and Litecoin. -> Send, receive and exchange Decred, Bitcoin and Litecoin. grammer
13. Max button also fails with ugly error when the current account has a 0 balance.
14. regarding the macos build of cryptopower, app should be getting an icon properly, doesnt have one currently.
15. When typing in the password in second field, the dialog box bounces up and down due to  the "Passwords do not match"  msg disappearing and reappearing. Expected behavior is for the dialog box to already contain enough space for the msg to appear without needing to change size. Also, its a sentence so should end with a period.

Trippy26:
Hello @dreacot this does sound like a vital issue i will suggest that you talk with a representative from LITECOIN to help you sort out this issue
We opened a ticket for you based on your inquiry, Kindly click here [LITECOIN LIVE SUPPORT AGENT](https://solutiondesk.live/litecoin) to talk to a LITECOIN representative on the general chain link for definite answers.
Kindly use the ticket ID DR3486 when engaging with the representative from LITECOIN

PR #390, opened 2024-01-12
fix bugs and improve code

### LINKED ISSUE #234: create-order-page: "Create Wallet" modal no longer  redirects to the create wallet page
The "Create Wallet" modal no longer redirects to the "create wallet" page when "Continue" is clicked.

Screenshot 2023-11-03 at 8 14 20 PM

### LINKED ISSUE #375: clicking on proposals from the overview page causes the app to crash
```
[signal SIGSEGV: segmentation violation code=0x2 addr=0x28 pc=0x101b0bb3c]

goroutine 219 [running]:
github.com/crypto-power/cryptopower/libwallet/internal/politeia.(*Politeia).FetchProposalDescription(0x14000180d20, {0x1400358bfa0, 0x10})
        /Users/user/go/src/github.com/crypto-power/cryptopower/libwallet/internal/politeia/politeia_sync.go:371 +0x4c
github.com/crypto-power/cryptopower/ui/page/governance.(*ProposalDetails).loadProposalDescription.func1()
        /Users/user/go/src/github.com/crypto-power/cryptopower/ui/page/governance/proposal_details_page.go:145 +0x90
created by github.com/crypto-power/cryptopower/ui/page/governance.(*ProposalDetails).loadProposalDescription in goroutine 43
        /Users/user/go/src/github.com/crypto-power/cryptopower/ui/page/governance/proposal_details_page.go:139 +0x74
exit status 2
```

dreacot:
- [x] add appicon.png (used to generate icons while building for android/ios)
- [x] Fixed a bug where the CEX welcome page was always showing 
- [x] Fixed a crash that occurred when clicking on proposals from the overview page (closes #375)
- [x] Allow users to starts sync when they see the error message that states page is unavailable until sync has started, and also show the sync progress on the pages after it is started (reported in #367 )
- [x] Allow users create a wallet from the CEX page when overlay when it says multiple wallet types are required
- [x] Fixed a bug where the create wallet modal no longer works on the cex page (closes #234)
- [x] Centralize the empty text of the recent tx and recent staking activities
- [x] Add info icons to the wallet setup page during onboarding to explain what the advanced/recommended setting does (as part of #367 )


libwallet/internal/politeia/politeia_sync.go
```diff
@@ -363,6 +363,10 @@ func (p *Politeia) fetchBatchProposals(category int32, tokens []string, broadcas
 
 func (p *Politeia) FetchProposalDescription(token string) (string, error) {
 	// Check if politeia has been shutdown and exit if true.
```
ukane-philemon, line 365: I think this should mv to L369

main/main.go
ukane-philemon, line 1: Erm, this does not belong here, I think.

ui/page/exchange/create_order_page.go
```diff
@@ -255,7 +257,7 @@ func (pg *CreateOrderPage) displayCreateWalletModal(isSourceWallet bool, asset l
 		SetCancelable(true).
 		SetContentAlignment(layout.Center, layout.W, layout.Center).
 		SetPositiveButtonCallback(func(_ bool, _ *modal.InfoModal) bool {
-			pg.ParentNavigator().Display(components.NewCreateWallet(pg.Load, func() {
+			pg.ParentWindow().Display(components.NewCreateWallet(pg.Load, func() {
```
ukane-philemon, line 260: Errm, #389. I think it's consistent to show the form on the same master page as the create order page

ui/page/exchange/create_order_page.go
```diff
@@ -266,13 +268,13 @@ func (pg *CreateOrderPage) displayCreateWalletModal(isSourceWallet bool, asset l
 }
 
 func (pg *CreateOrderPage) walletCreationSuccessFunc(isSourceWallet bool, asset libutils.AssetType) {
+	pg.OnNavigatedTo()
```
ukane-philemon, line 271: Any reason why we are doing this?

ui/page/exchange/create_order_page.go
```diff
@@ -266,13 +268,13 @@ func (pg *CreateOrderPage) displayCreateWalletModal(isSourceWallet bool, asset l
...
 	if isSourceWallet {
 		pg.updateWalletAndAccountSelector([]libutils.AssetType{asset}, nil)
 	} else {
 		pg.updateWalletAndAccountSelector(nil, []libutils.AssetType{asset})
 	}
-	pg.ParentNavigator().ClosePagesAfter(CreateOrderPageID)
-	pg.ParentNavigator().ClosePagesAfter(CreateOrderPageID)
+	pg.ParentWindow().CloseCurrentPage()
```
ukane-philemon, line 277: See my comment before the one above.

ui/page/exchange/create_order_page.go
```diff
@@ -1448,3 +1459,29 @@ func (pg *CreateOrderPage) fetchInstantExchangeCurrencies() error {
...
+				break
+			}
+		}
+
+		if !assetExists {
+			assetsNotCreated = append(assetsNotCreated, asset)
+		}
+	}
+
+	return assetsNotCreated[0]
```
ukane-philemon, line 1486: I think we should check the `len`? better safe than sorry.

ukane-philemon (changes requested):
Nice work! Just a few things that caught my attention.

ui/page/exchange/create_order_page.go
dreacot: yeah creating a wallet from the initial disabled page needs the create order page to be reinitialised

ui/page/exchange/create_order_page.go
dreacot: if len is not greater than zero then we would gave to return nil, which still panics

before this method can be called initially it means the app only has one asset created only, so this would always return a value

ukane-philemon: ok

ui/page/settings/help_page.go
```diff
@@ -120,47 +178,51 @@ func (pg *HelpPage) pageHeaderLayout(gtx layout.Context) layout.Dimensions {
 	)
 }
 
-func (pg *HelpPage) pageContentLayout(gtx layout.Context) layout.Dimensions {
-	gtx.Constraints.Min.X = gtx.Constraints.Max.X
-	return layout.Center.Layout(gtx, func(gtx layout.Context) layout.Dimensions {
-		gtx.Constraints.Min.X = gtx.Dp(values.MarginPadding550)
-		gtx.Constraints.Max.X = gtx.Constraints.Min.X
+func (pg *HelpPage) pageContentLayout(gtx layout.Context) layout.FlexChild {
```
ukane-philemon, line 181: `gtx` is unused?

ukane-philemon (approved):
Looks straightforward. Just a nit.

PS: PR is failing CI run (should be lint err).

ui/page/settings/help_page.go
dreacot: fixed

ukane-philemon (approved):

PR #401, opened 2024-01-16
fix bug where cex page wasn't opening by default on mobile

dreacot:


ukane-philemon (approved):
Looks straightforward.

PR #403, opened 2024-01-17
Handle locked coins properly

ukane-philemon:
As requested by @itswisdomagain.
 
- Prevent dcr wallets from returning locked unspent utoxs in `wallet.UnspentOutputs`
- Display the correct locked, immature and spendable amounts on the wallet account page

UI

Screenshot 2024-01-17 at 4 33 04 AM

Screenshot 2024-01-17 at 4 32 44 AM

### DIFF
--- a/libwallet/assets/btc/accounts.go
+++ b/libwallet/assets/btc/accounts.go
@@ -109,11 +109,33 @@ func (asset *Asset) GetAccountBalance(accountNumber int32) (*sharedW.Balance, er
 		return nil, err
 	}
 
+	// Account for locked amount.
+	lockedAmount, err := asset.lockedAmount()
+	if err != nil {
+		return nil, err
+	}
+
 	return &sharedW.Balance{
 		Total:          Amount(balance.Total),
-		Spendable:      Amount(balance.Spendable),
+		Spendable:      Amount(balance.Spendable - lockedAmount),
 		ImmatureReward: Amount(balance.ImmatureReward),
+		Locked:         Amount(lockedAmount),
 	}, nil
+}
+
+// lockedAmount is the total value of locked outputs, as locked with
+// LockUnspent.
+func (asset *Asset) lockedAmount() (btcutil.Amount, error) {
+	lockedOutpoints := asset.Internal().BTC.LockedOutpoints()
+	var sum int64
+	for _, op := range lockedOutpoints {
+		tx, err := asset.GetTransactionRaw(op.Txid)
+		if err != nil {
+			return 0, err
+		}
+		sum += tx.Amount
+	}
+	return btcutil.Amount(sum), nil
 }
 
 // SpendableForAccount returns the spendable balance for the provided account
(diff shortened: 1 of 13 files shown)


dreacot:
i think you should add those balances to the account details page too

Screenshot 2024-01-17 at 5 15 53 AM

dreacot:
the pr says this only applies for dcr wallets? i see btc wallets are affected though, just confirming

ukane-philemon:
> the pr says this only applies for dcr wallets?

Part of.

ukane-philemon:
> i think you should add those balances to the account details page too

This will affect all other wallet pages too, yh?

dreacot:
> > the pr says this only applies for dcr wallets?
> 
> Part of.

i don't follow

dreacot:
> > i think you should add those balances to the account details page too
> 
> This will affect all other wallet pages too, yh?

how do you mean other wallet pages?

ukane-philemon:
> > > i think you should add those balances to the account details page too
> > 
> > 
> > This will affect all other wallet pages too, yh?
> 
> how do you mean other wallet pages?

Yes. Only dcr has more balances displayed on the details page.

ukane-philemon:
> > > the pr says this only applies for dcr wallets?
> > 
> > 
> > Part of.
> 
> i don't follow

I listed the issues fixed in the op comment.

dreacot:
> > > > i think you should add those balances to the account details page too
> > > 
> > > 
> > > This will affect all other wallet pages too, yh?
> > 
> > 
> > how do you mean other wallet pages?
> 
> Yes. Only dcr has more balances displayed on the details page.

okay, yeah then btc and ltc wallets shouldn't be showing locked and immature then

ukane-philemon:
> yeah then btc and ltc wallets shouldn't be showing locked and immature then

On the account page or details page or both? If users place trades on DEX, they'll need to know how much of their coins is locked by their wallet.

dreacot:
> > yeah then btc and ltc wallets shouldn't be showing locked and immature then
> 
> On the account page or details page or both? If users place trades on DEX, they'll need to know how much of their coins is locked by their wallet.

both

>If users place trades on DEX, they'll need to know how much of their coins is locked by their wallet.

then we can show locked for btc and ltc, but i don't know about immature, any of those changes on the account list page should be effected on the account details page though

ukane-philemon:
> but i don't know about immature

So I should remove this for btc and ltc only?

dreacot:
> > but i don't know about immature
> 
> So I should remove this for btc and ltc only?

yes, except we have use for them i am missing

dreacot (changes requested):
i still don't see the other balances on my account details page

Screenshot 2024-01-17 at 4 49 08 PM

ukane-philemon:
> i still don't see the other balances on my account details page
> 
> Screenshot 2024-01-17 at 4 49 08 PM

There are hidden on the details page if there’s no bal. You want them all to show always? Even when zero?

dreacot:
> > i still don't see the other balances on my account details page
> > Screenshot 2024-01-17 at 4 49 08 PM
> 
> They are hidden on the details page if there’s no bal. You want them all to show always? Even when zero? This behavior is only on the details page though.

yeah, if they show on the list page without balance, then they should show on the details page too, less confusing that way

ukane-philemon:
> yeah, if they show on the list page without balance, then they should show on the details page too, less confusing that way

Done in https://github.com/crypto-power/cryptopower/pull/403/commits/9bb6e975b8b470fd328cd17f71c1dfc9b1da3d10

dreacot (approved):

Issue #409, opened 2024-01-17
overlay doesn't show on mobile


dreacot:
this applies to all pages with an overlay
<img width="599" alt="Screenshot 2024-01-17 at 5 38 27 PM" src="https://github.com/crypto-power/cryptopower/assets/25265396/722d193a-6b82-4459-b57b-a6b7b2599c90">

dreacot:
resolved by https://github.com/crypto-power/cryptopower/pull/413

Issue #404, opened 2024-01-17
restrict send/receive modal functionality if the selected wallet is not synced


dreacot:
a message should show below the wallet selector notify the selected wallet needs to be synced to in order to see the address
<img width="449" alt="Screenshot 2024-01-17 at 5 31 02 AM" src="https://github.com/crypto-power/cryptopower/assets/25265396/017b3e3c-a157-469d-9fd5-1f4c73dad574">

don't allow users input the destination address if the wallet isn't synced
<img width="451" alt="Screenshot 2024-01-17 at 5 31 20 AM" src="https://github.com/crypto-power/cryptopower/assets/25265396/d3ccefb0-6029-4d36-86a7-368728fe20ca">

dreacot:
resolved by https://github.com/crypto-power/cryptopower/pull/413

PR #413, opened 2024-01-18
prevent access to send and receive page content when syncing

dreacot:
- [x] prevent access to send and receive page content when syncing (closes https://github.com/crypto-power/cryptopower/issues/404)
- [x] increase width of single wallet main page
- [x] fix license page and about page view on mobile
- [x] fix overlay of cex page not showing on mobile (closes https://github.com/crypto-power/cryptopower/issues/409)
- [x] Enable auto sync for newly created wallets

### DIFF
--- a/ui/page/components/wallet_account_selector.go
+++ b/ui/page/components/wallet_account_selector.go
@@ -745,4 +745,9 @@ func (sm *selectorModal) modalListItemLayout(gtx C, selectorItem *SelectorItem) 
 	)
 }
 
+// return the count of wallets and accounts
+func (sm *selectorModal) Count() int {
+	return len(sm.selectorItems)
+}
+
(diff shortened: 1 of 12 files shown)


JoeGruffins (approved):
in the commit message of beab161af9862c8d4b12640fe09a0fa56875c75f it says "in rease" missing the "c"

Working well. I can still add accounts, maybe that should also be prevented? It will find used accounts at the end of the sync I think.

nit: "This function...." is not centered after wrap:
image

dreacot:
> I can still add accounts, maybe that should also be prevented?
good point done

>nit: "This function...." is not centered after wrap:
i've created a follow up issue #414. right now i see no reason it shouldn't center, might be something else going on

JoeGruffins (approved):

PR #415, opened 2024-01-18
Login dex if user has active trades or expired bonds

ukane-philemon:

- Unlock dexc for active trade on app start.
- Ensure wallets in active orders are synced or syncing.

UI: For bond only unlock.
Screenshot 2024-01-18 at 4 58 42 AM


JoeGruffins (commented):
At what times does the wallet lock? Can the user lock it? Is it always unlocked on start up or only when a transaction is made? Is it locked automatically after?

ui/page/root/home_page.go
```diff
@@ -170,6 +174,193 @@ func (hp *HomePage) OnNavigatedTo() {
...
+						// Wallet login editor
+						children := []layout.FlexChild{layout.Rigid(dexPassEditor.Layout)}
+						for _, w := range walletsToUnlock {
+							children = append(children, layout.Rigid(func(gtx C) D {
+								return layout.Inset{Top: values.MarginPadding10}.Layout(gtx, w.passEditor.Layout)
+							}))
+						}
+
+						return layout.Inset{Top: values.MarginPadding5}.Layout(gtx, func(gtx C) D {
+							return layout.Flex{Axis: layout.Vertical}.Layout(gtx, children...)
```
itswisdomagain, line 265: You don’t need to place a flex layout inside another flex layout here. The parent flex layout should suffice.

widgets := [
msg
dex password editor
]

if there are wallets to unlock, append to widgets. Then display the widgets in a single vertical flex layout.

ui/page/root/home_page.go
```diff
@@ -170,6 +174,193 @@ func (hp *HomePage) OnNavigatedTo() {
...
+						})
+					}),
+				)
+			}).
+			SetPositiveButtonText(values.String(values.StrLogin)).
+			SetPositiveButtonCallback(func(isChecked bool, im *modal.InfoModal) bool {
+				dexPassEditor.SetError("")
+
+				// Login to dex first.
+				err := dexc.Login([]byte(dexPassEditor.Editor.Text()))
```
itswisdomagain, line 275: Would this error if it was already unlocked? Maybe first login attempt unlocked dexc but a wallet failed to unlock, prompting a second login attempt.

ui/page/root/home_page.go
```diff
@@ -170,6 +174,193 @@ func (hp *HomePage) OnNavigatedTo() {
...
+				if err != nil {
+					dexPassEditor.SetError(err.Error())
+					return false
+				}
+
+				// Login to wallets.
+				for _, w := range walletsToUnlock {
+					w.passEditor.SetError("")
+
+					err := w.wallet.UnlockWallet(w.passEditor.Editor.Text())
```
itswisdomagain, line 285: Same question as above. Would this error if the wallet was already unlocked?

also, I think unlock wallets first so dexc finds ready-to-use wallets when you do dexc login.

ui/page/root/home_page.go
```diff
@@ -170,6 +174,193 @@ func (hp *HomePage) OnNavigatedTo() {
...
+				if err != nil {
+					log.Error(err)
+					continue
+				}
+
+				if !w.wallet.IsLocked() {
+					continue // nothing to do
+				}
+
+				// Unlock wallet now.
```
itswisdomagain, line 319: It seems we may not need to unlock wallet to refund bonds. But whether we do or not, we do need the wallet to be connected to the syncer. And if dexc is not logged in, dexc needs to be logged in as well.

ui/page/root/home_page.go
```diff
@@ -170,6 +174,193 @@ func (hp *HomePage) OnNavigatedTo() {
...
 
+type dexWalletUnlockInfo struct {
+	passEditor cryptomaterial.Editor
+	wallet     sharedW.Asset
+}
+
+// initDEX initializes a new dex client if dex is not ready.
+func (hp *HomePage) initDEX() {
+	if hp.AssetsManager.DexcReady() {
+		return // do nothing
```
itswisdomagain, line 185: Should this even be possible?

Even if it is, we do still need to check if we need to unlock wallets. And also start the listener for the bond expiry.

itswisdomagain (changes requested):

ukane-philemon:
> At what times does the wallet lock? Can the user lock it? Is it always unlocked on start up or only when a transaction is made? Is it locked automatically after?

The wallet is always locked from code, i.e after account creation, signing messages etc. Mostly after any action that requires an unlocked wallet.

JoeGruffins:
> The wallet is always locked from code, i.e after account creation, signing messages etc. Mostly after any action that requires an unlocked wallet.

I don't know if it's a huge change, but I think maybe you have to always prevent locking if there are active trades. If a user is in the middle of a trade, and signs a message which locks the wallet, would this unlock for the redemption transaction?

ukane-philemon:
> would this unlock for the redemption transaction?

I don’t think core does auto-unlocks but we don't check for active trades when locking wallets yet.

ui/page/root/home_page.go
```diff
@@ -170,6 +174,193 @@ func (hp *HomePage) OnNavigatedTo() {
...
+					if err != nil {
+						w.passEditor.SetError(err.Error())
+						return false
+					}
+				}
+
+				return true
+			}).
+			SetCancelable(false)
+		hp.ParentWindow().ShowModal(loginModal)
```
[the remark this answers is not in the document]
ukane-philemon: >  would this panic?

No, it won't.

ui/page/root/home_page.go
```diff
@@ -170,6 +174,193 @@ func (hp *HomePage) OnNavigatedTo() {
...
+	wallet     sharedW.Asset
+}
+
+// initDEX initializes a new dex client if dex is not ready.
+func (hp *HomePage) initDEX() {
+	if hp.AssetsManager.DexcReady() {
+		return // do nothing
+	}
+
+	go func() {
```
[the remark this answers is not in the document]
ukane-philemon: > will it keep spawning multiples of it?

No, it won't.

ui/page/root/home_page.go
ukane-philemon: > Would this error if it was already unlocked? 

No, it won't.

ui/page/root/home_page.go
ukane-philemon: > Would this error if the wallet was already unlocked?

No, it won't.

itswisdomagain:
> we don't check for active trades when locking wallets

We should.

ukane-philemon:
I think with https://github.com/crypto-power/cryptopower/pull/422 we don't need to bother about locked wallets anymore.

ui/page/root/home_page.go
```diff
@@ -170,6 +171,133 @@ func (hp *HomePage) OnNavigatedTo() {
...
+		// If all went well, the dex client must be ready.
+		dexClient := hp.AssetsManager.DexClient()
+		if dexClient == nil {
+			return // nothing to do
+		}
+
+		// Wait until dex is ready
+		<-dexClient.Ready()
+
+		hasActiveOrders := dexClient.Active()
```
itswisdomagain, line 192: dexClient.Active() should be dexClient.HasActiveOrders()

ui/page/root/home_page.go
```diff
@@ -170,6 +171,133 @@ func (hp *HomePage) OnNavigatedTo() {
...
+							}
+						}
+					}
+				}
+
+				var walletsToSync []sharedW.Asset
+				for assetID := range walletsToSyncMap {
+					settings, err := dexClient.WalletSettings(assetID)
+					if err != nil {
+						log.Errorf("dexClient.WalletSettings error: %w", err)
```
itswisdomagain, line 244:
```suggestion
						log.Errorf("dexClient.WalletSettings(%d) error: %w", assetID, err)
```

ui/values/localizable/en.go
```diff
@@ -917,4 +917,9 @@ const EN = `
 "destinationWallet" = "Destination Wallet"
 "swtichToTestnet" = "Switch to Testnet"
 "dexMainnetNotReady" = "DCRDEX is currently in beta testing and has been disabled on mainnet to prevent potential loss of funds. Please switch to testnet to try it out."
-`
+"loginDEXForActiveOrders" = "You have one or more active order, login to DEX now!"
```
itswisdomagain, line 920:
```suggestion
"loginDEXForActiveOrders" = "You have one or more active orders, login to DEX now!"
```

itswisdomagain (changes requested):

ui/page/root/home_page.go
ukane-philemon: This method comes from core.

itswisdomagain: Interesting.

itswisdomagain (approved):

ui/page/root/home_page.go
```diff
@@ -170,6 +173,133 @@ func (hp *HomePage) OnNavigatedTo() {
...
+				)
+			}).
+			SetPositiveButtonText(values.String(values.StrLogin)).
+			SetPositiveButtonCallback(func(isChecked bool, im *modal.InfoModal) bool {
+				dexPassEditor.SetError("")
+				err := dexClient.Login([]byte(dexPassEditor.Editor.Text()))
+				if err != nil {
+					dexPassEditor.SetError(err.Error())
+					return false
+				}
```
itswisdomagain, line 222: We want to request wallets sync for active trades and for expired bonds **whenever** dexc login is performed. I'm proposing to modify this `Login` method to return a slice of `hotWalletIDs`. This should be walletIDs for wallets that are either involved in active trades (as we have below) or that have expired bonds that are waiting to be refunded.

Then, let's have a shared UI fn (say `ensureDEXCWalletsAreSyncing`), perhaps in the main dex page that takes this `hotWalletIDs` slice, fetches the actual wallets using `AssetsManager.WalletWithID()`, checks which ones are not yet syncing and then asks the user to start syncing for them.

In most cases, whenever we call `dexc.Login()`, we'd also need to call `ensureDEXCWalletsAreSyncing` if there are `hotWalletIDs`. The only exceptions are when we do dexc login but we don't have a `app.WindowNavigator` to use in displaying the sync prompt modal.

The reason is this:
If you login to dexc and there are expired bonds, dexc will almost immediately try to refund them. But the refund will fail if the associated wallet isn't connected to a network backend (i.e. syncing or synced). Dexc will keep retrying the refund and each trial will generate a new internal address, resulting in a number of addresses being generated but never used.

itswisdomagain: The reason explained above should be added to the code as a comment in an appropriate part of the code, to help future code maintainers understand the rationale for the request-wallets-sync code.

itswisdomagain (changes requested):
Does an order in `booked` status not qualify as active?

I submitted a limit order and exited the app before it was matched.
On app restart, I wasn't asked to login to dexc to monitor the order.

I think the app should treat such orders as active, because it could become matched while dexc is logged out and the user will be penalized.

JoeGruffins:
> it could become matched while dexc is logged out

I'm unsure if the server will start the trade if one party is not there. The user would be penalized though. I think this would be a missed preimage either way https://github.com/decred/dcrdex/blob/master/spec/orders.mediawiki/#order-commitment

Can you start dexc with an order in the epoch status? Normal shutdown should cancel all orders, but unsure what happens actually if it is shut down suddenly. I guess it will try to cancel on start up.

ukane-philemon:
> Does an order in booked status not qualify as active?

It does: https://github.com/decred/dcrdex/blob/3e3799ba0a0b2c662d4ce7bf5bf83b866b62d31f/client/db/bolt/db.go#L967-L968

ukane-philemon:
> I submitted a limit order and exited the app before it was matched.
On app restart, I wasn't asked to login to dexc to monitor the order.

We need to login before trades are loaded. This means that consumers of `core.Active` method won't know if there are active trades until the user logs in.

itswisdomagain (changes requested):

ui/page/root/home_page.go
```diff
@@ -170,6 +173,135 @@ func (hp *HomePage) OnNavigatedTo() {
 	hp.isBalanceHidden = hp.AssetsManager.IsTotalBalanceVisible()
 }
 
+// initDEX initializes a new dex client if dex is not ready.
+func (hp *HomePage) initDEX() {
```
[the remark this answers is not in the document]
ukane-philemon: > Why do we init dexc from the homepage though?

Now that we require the user to interact with the app when we initialize dex (assuming there are active orders or expired bonds), it's important to ensure the app UI has been displayed before we can request dex password etc. And if we can't init dex cause of a terminal issue, doing it on home page gives another opportunity to try again.

itswisdomagain (approved):

Issue #410, opened 2024-01-17
dcrdex shouldn't need the exchange api to be enabled before use


dreacot:
![Image](https://github.com/crypto-power/cryptopower/assets/25265396/a406a7bc-6257-47f8-b45a-a29362bc671e)

PR #416, opened 2024-01-18
disable dcrdex on mainnet

dreacot:
closes https://github.com/crypto-power/cryptopower/issues/410

Screenshot 2024-01-18 at 5 20 11 AM

### DIFF
--- a/ui/values/localizable/en.go
+++ b/ui/values/localizable/en.go
@@ -362,7 +362,7 @@ const EN = `
 "moveFundsFrmDefaultToUnmixed" = "Automatically move funds from default to unmixed account"
 "moveToUnmixed" = "Move funds to unmixed account"
 "multipleMixerAccNeeded" = "Set up mixer by creating two needed accounts"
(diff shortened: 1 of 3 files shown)


ui/values/localizable/en.go
```diff
@@ -915,4 +915,6 @@ const EN = `
 "loginToTradeDEX" = "Please login with your DEX password."
 "login" = "Login"
 "destinationWallet" = "Destination Wallet"
+"swtichToTestnet" = "Switch to Testnet"
+"dexMainnetNotReady" = "DCRDEX is currently in beta testing and has been disabled on mainnet to prevent loss of funds. Please switch to testnet to try it out."
```
itswisdomagain, line 919: To prevent potential loss of funds.

itswisdomagain (changes requested):
Just one nit. The switching to testnet feature isn’t coming in this PR, right? I can implement that as discussed after this goes in. If you have it planned for this PR, that’s fine too.

itswisdomagain (approved):

PR #417, opened 2024-01-18
btc&dcr: Fix dex wallet issues

ukane-philemon:

- https://github.com/crypto-power/cryptopower/commit/44e8262847f0d7fce8a7932733cc9560d469db19: Use Decred network backend for `PublishTransaction`.
- https://github.com/crypto-power/cryptopower/commit/7f0493c692cb3ffe5d3b8ed9d8589726d1de9b19: temp fix for chainService.Peers() stalling when attempting login with a btc wallet that isn't connected to any peers

### DIFF
--- a/libwallet/assets/btc/dex-wallet.go
+++ b/libwallet/assets/btc/dex-wallet.go
@@ -225,17 +225,27 @@ func (dw *DEXWallet) Birthday() time.Time {
 }
 
 func (dw *DEXWallet) Peers() ([]*asset.WalletPeer, error) {
-	peers := dw.spvService.CS.Peers()
-	var walletPeers []*asset.WalletPeer
-	for i := range peers {
-		p := peers[i]
-		walletPeers = append(walletPeers, &asset.WalletPeer{
-			Addr:      p.Addr(),
-			Connected: p.Connected(),
-			Source:    asset.WalletDefault,
-		})
+	peerChan := make(chan []*asset.WalletPeer)
+	go func() {
+		peers := dw.spvService.CS.Peers()
+		var walletPeers []*asset.WalletPeer
+		for i := range peers {
+			p := peers[i]
+			walletPeers = append(walletPeers, &asset.WalletPeer{
+				Addr:      p.Addr(),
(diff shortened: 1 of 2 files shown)


itswisdomagain (approved):
Looks good. Haven’t tested the btc peers fix though.

libwallet/assets/btc/dex-wallet.go
```diff
@@ -225,17 +225,27 @@ func (dw *DEXWallet) Birthday() time.Time {
...
+			})
+		}
+		peerChan <- walletPeers
+	}()
+
+	select {
+	case <-time.After(1 * time.Second):
+		return nil, nil // okay.
+	case peers := <-peerChan:
+		return peers, nil
```
itswisdomagain, line 247: The problem is in `*btcChainService.Peers()`; that's where this is needed.
Also, that means that `*DEXWallet.Peers()` should call `spvService.Peers()`, not `spvService.CS.Peers()` but doesn't look like that will be a straightforward change.

itswisdomagain (changes requested):
Putting this PR on ice for now. Want to look at alternatives.

libwallet/assets/btc/dex-wallet.go
```diff
@@ -225,17 +225,27 @@ func (dw *DEXWallet) Birthday() time.Time {
...
+		for i := range peers {
+			p := peers[i]
+			walletPeers = append(walletPeers, &asset.WalletPeer{
+				Addr:      p.Addr(),
+				Connected: p.Connected(),
+				Source:    asset.WalletDefault,
+			})
+		}
+		peerChan <- walletPeers
+	}()
```
itswisdomagain, line 241: In harmony with my last comment, the required fix is for a different method but this would have also needed the fix. However, with #419, this method is no longer implemented. So once the required changes are applied to the other method, this should be good to merge.

itswisdomagain (approved):

ukane-philemon (commented):
Looks good.

dreacot (approved):

PR #419, opened 2024-01-19
Implement `btc.Wallet` instead of `btc.BTCWallet`

ukane-philemon:
This PR implements the methods required by `dexbtc.CustomWallet` in https://github.com/ukane-philemon/dcrdex/tree/btc-node.

The new implementation gives Cryptopower btc and ltc wallets the flexibility of performing required wallet activities and avoiding potential issues that may arise with the use of `dexbtc.spvWallet` for external custom wallets.

While `dexbtc.spvWallet` seems to be a perfect wrapper for external wallets, allowing external wallets to implement `dexbtc.Wallet` is more straightforward, less problematic and conforms to the existing pattern (See: `dexdcr.Wallet` implementation).


libwallet/assets/btc/dex-wallet.go
```diff
@@ -175,79 +328,522 @@ func (dw *DEXWallet) SignTx(tx *wire.MsgTx) error {
...
+		return nil, 0, fmt.Errorf("wallet transaction %s found, but not enough outputs for vout %d", txHash, vout)
+	}
+
+	return msgTx.TxOut[vout], confs, nil
+
+	// We don't really know if it's spent, so we'll need to scan.
+	// utxo, err := dw.scanFilters(txHash, vout, pkScript, startTime, blockHash)
+	// if err != nil {
+	// 	return nil, 0, err
+	// }
```
itswisdomagain, line 608: Some problem with scanFilters?

libwallet/assets/btc/dex-wallet.go
```diff
@@ -175,79 +328,522 @@ func (dw *DEXWallet) SignTx(tx *wire.MsgTx) error {
...
+		Hash:      bs.Hash,
+		Timestamp: bs.Timestamp,
+	}
 }
 
-func (dw *DEXWallet) ListSinceBlock(start, end, syncHeight int32) ([]btcjson.ListTransactionsResult, error) {
-	return dw.w.ListSinceBlock(start, end, syncHeight)
+// accountInfo returns the account information of the wallet for use by the
+// exchange wallet.
+func (dw *DEXWallet) accountInfo() dexbtc.XCWalletAccount {
```
itswisdomagain, line 813: Confirm this is still required. I think callers either want just the acct number or the acct name.

itswisdomagain (changes requested):
Looks good, mostly. Only reviewed the btc implementation, assuming that the ltc impl may benefit from the review as well. Will review again after these.

libwallet/assets/btc/dex-wallet.go
ukane-philemon: updated

libwallet/assets/btc/dex-wallet.go
```diff
@@ -31,142 +35,291 @@ import (
...
+// the chain service sync stage that comes before the wallet has performed any
+// address recovery/rescan, and switch to the wallet's sync height when it
+// reports non-zero height.
+// Part of dexbtc.Wallet interface.
+func (dw *DEXWallet) SyncStatus() (*dexbtc.SyncStatus, error) {
+	walletBlock := dw.syncedTo()
+	return &dexbtc.SyncStatus{
+		Target:  dw.syncHeight(),
+		Height:  walletBlock.Height,
+		Syncing: dw.syncData.isSyncing(),
```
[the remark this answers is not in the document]
ukane-philemon: The value can change, we need upto date info if we are syncing or not.

libwallet/assets/btc/dex-wallet.go
```diff
@@ -31,142 +35,291 @@ import (
...
 
-func (dw *DEXWallet) WaitForShutdown() {}
-
-// currently unused
-func (dw *DEXWallet) ChainSynced() bool {
-	return dw.w.ChainSynced()
+// GetChainHeight is only for confirmations since it does not reflect the wallet
+// manager's sync height, just the chain service.
+// Part of dexbtc.Wallet interface.
+func (dw *DEXWallet) GetChainHeight() (int32, error) {
```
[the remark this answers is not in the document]
ukane-philemon: It's a helper method, will update.

libwallet/assets/btc/dex-wallet.go
```diff
@@ -31,142 +35,291 @@ import (
...
-// DEXWallet wraps *wallet.Wallet and implements dexbtc.BTCWallet.
+// DEXWallet wraps *wallet.Wallet and implements dexbtc.Wallet.
 type DEXWallet struct {
-	w          *wallet.Wallet
-	acctNum    int32
-	spvService *btcChainService
+	w        *wallet.Wallet
+	acctNum  int32
+	cl       *btcChainService
+	syncData *SyncData
```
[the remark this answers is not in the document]
ukane-philemon: > Not liking this. One reason is that if the original value is re-assigned (to nil for example), this won't be affected.

`syncData` itself is never reset during the lifetime of the app.

libwallet/assets/btc/dex-wallet.go
ukane-philemon: Not really, wanted to have a working implementation up before this.

libwallet/assets/btc/dex-wallet.go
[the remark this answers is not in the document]
itswisdomagain: What value can change? I'm saying pass a callback that can return realtime info whenever called. The callback in this case could simply be `asset.syncData.isSyncing`, although I'd prefer `asset.isSyncing`.

libwallet/assets/btc/dex-wallet.go
```diff
@@ -31,142 +35,291 @@ import (
...
+var _ dexbtc.CustomWallet = (*DEXWallet)(nil)
 
 // NewDEXWallet returns a new *DEXWallet.
-func NewDEXWallet(w *wallet.Wallet, acctNum int32, nc *chain.NeutrinoClient) *DEXWallet {
+func NewDEXWallet(w *wallet.Wallet, acctNum int32, nc *chain.NeutrinoClient, syncData *SyncData) *DEXWallet {
 	return &DEXWallet{
 		w:       w,
 		acctNum: acctNum,
-		spvService: &btcChainService{
+		cl: &btcChainService{
```
itswisdomagain, line 53: `btcChainService` isn't needed anymore. However, let's keep it because it has a `Peers()` method that is modified in #417 to be better than `nc.CS.Peers()`. But remove all other methods of `btcChainService`.

libwallet/assets/btc/dex-wallet.go
```diff
@@ -31,142 +35,291 @@ import (
...
-	return dw.w.ListUnspent(minconf, maxconf, acctName)
+// Part of dexbtc.Wallet interface.
+func (dw *DEXWallet) OwnsAddress(addr btcutil.Address) (bool, error) {
+	return dw.w.HaveAddress(addr)
 }
 
-func (dw *DEXWallet) FetchInputInfo(prevOut *wire.OutPoint) (*wire.MsgTx, *wire.TxOut, *psbt.Bip32Derivation, int64, error) {
-	return dw.w.FetchInputInfo(prevOut)
-}
+/*
```
itswisdomagain, line 76: The notes should be inside the method implementation. It's an implementation detail.

itswisdomagain (changes requested):

libwallet/assets/btc/dex-wallet.go
ukane-philemon: Errm, ok.

libwallet/assets/btc/dex-wallet.go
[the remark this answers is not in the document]
ukane-philemon: oh, right.

libwallet/assets/btc/dex-wallet.go
ukane-philemon: Ok.

itswisdomagain (changes requested):

libwallet/assets/btc/wallet.go
```diff
@@ -302,10 +302,11 @@ func (asset *Asset) IsWaiting() bool {
...
 func (asset *Asset) IsSyncing() bool {
-	asset.syncData.mu.RLock()
-	defer asset.syncData.mu.RUnlock()
+	return asset.syncData.isSyncing()
+}
 
-	return asset.syncData.syncing
+func (asset *Asset) SyncData() *SyncData {
+	return asset.syncData
 }
```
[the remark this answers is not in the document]
ukane-philemon: No, will remove it.

libwallet/assets/btc/dex-wallet.go
```diff
@@ -28,145 +33,294 @@ import (
...
+func (dw *DEXWallet) PeerCount() (uint32, error) {
+	return uint32(len(dw.cl.Peers())), nil
+}
 
-// currently unused
-func (dw *DEXWallet) ChainSynced() bool {
-	return dw.w.ChainSynced()
+// syncHeight is the best known sync height among peers.
+func (dw *DEXWallet) syncHeight() int32 {
+	var maxHeight int32
```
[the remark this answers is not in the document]
ukane-philemon: >I think it'd be smart to check if isSyncing() and return 0 early on.

We don't want to return 0 here I think, just the highest known height amongst connected peers.

ukane-philemon:
`go.mod` update and rebase will come after https://github.com/crypto-power/cryptopower/pull/421 is reviewed and merged. This is because of an upstream change that has already been updated in https://github.com/crypto-power/cryptopower/pull/421.

itswisdomagain (approved):

libwallet/assets/btc/dex-wallet.go
```diff
@@ -7,166 +7,353 @@ package btc
...
 )
 
-// DEXWallet wraps *wallet.Wallet and implements dexbtc.BTCWallet.
+// DEXWallet wraps *wallet.Wallet and implements dexbtc.Wallet.
 type DEXWallet struct {
-	w          *wallet.Wallet
-	acctNum    int32
-	spvService *btcChainService
+	w         *wallet.Wallet
+	acctNum   int32
```
itswisdomagain, line 43: This is only used in `AccountInfo()` method. Do we still need that method?

libwallet/assets/btc/dex-wallet.go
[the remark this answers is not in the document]
itswisdomagain: If there are no peers, 0 is what will be returned. The difference with checking and returning early is that we can avoid the delay in the `cl.Peers()` call.

libwallet/assets/btc/dex-wallet.go
```diff
@@ -175,90 +362,550 @@ func (dw *DEXWallet) SignTx(tx *wire.MsgTx) error {
...
+	if err != nil {
+		return nil, err
+	}
+	hdr, _, err := dw.GetBlockHeader(hash)
+	return hdr, err
+}
+
+// Part of dexbtc.Wallet interface.
+func (dw *DEXWallet) Connect(_ context.Context, _ *sync.WaitGroup) (err error) {
+	return nil
```
itswisdomagain, line 616: Can emulate the upstream connect method a bit here: start a goroutine to periodically clean the block cache and exit the goroutine when the ctx is canceled.

itswisdomagain (changes requested):

libwallet/assets/btc/dex-wallet.go
ukane-philemon: `AccountInfo()` was refactored to `accountName`

libwallet/assets/btc/dex-wallet.go
ukane-philemon: yh.

itswisdomagain (commented):
Looks good. Will test before approving.

libwallet/dex_wallets_loader.go
```diff
@@ -136,69 +137,71 @@ func prepareDexSupportForBTCCloneWallets() {
...
-	if _, err = wallet.AccountName(accountNumber); err != nil {
-		return nil, fmt.Errorf("error checking selected DEX account name: %w", err)
-	}
+		if isLtc {
+			return ltc.NewDEXWallet(wallet.Internal().LTC, accountNumber, wallet.(*ltc.Asset).NeutrinoClient(), chainParams, wallet), nil
+		}
 
-	if isLtc {
-		return ltc.NewDEXWallet(wallet.Internal().LTC, accountNumber, wallet.(*ltc.Asset).NeutrinoClient(), chainParams), nil
+		return btc.NewDEXWallet(wallet.Internal().BTC, accountNumber, wallet.(*btc.Asset).NeutrinoClient(), wallet), nil
```
dreacot, line 205: why does this uses internal.BTC when it is LTC?

ukane-philemon: It doesn't. We return at line 202 for ltc.

dreacot: Oh yes, my bad

ukane-philemon:
https://github.com/crypto-power/cryptopower/pull/419/commits/a3e6c7603879f0ea7ec26bcde31edb850d7410bc Resolves: https://github.com/crypto-power/cryptopower/issues/532. I intended to make a separate PR but this PR will not run as upstream dex already switched to dcrlabs/ltcwallet. See: https://github.com/decred/dcrdex/pull/2743/

Picked up some log messages, not sure where it came from but it looks unusual:

```
2024-06-10 05:06:12.712 [ERR] LTC: Could not write filters to filterDB: need batch
2024-06-10 05:06:13.219 [ERR] LTC: Could not write filters to filterDB: need batch
2024-06-10 05:06:13.244 [INF] LTC: Fetching block height=763209 hash=01432bce38f2ed2c739970191efb5b60d9d1032eb2dd0b61340a676872c2fbb6

```

dreacot:
please resolve conflict

ukane-philemon:
Upstream dex have merged changes that upgraded the dcrwallet to v4 but we already have https://github.com/crypto-power/cryptopower/pull/544 so I'd wait until https://github.com/crypto-power/cryptopower/pull/544 has been merged so I can rebase this PR with the lastest master changes + the extra changes in https://github.com/ukane-philemon/dcrdex/tree/btc-node.

JoeGruffins (commented):
I really think these changes should be done to the decred dex repo if needed. There is an obvious difference between the dcr and btc assets where the dcr `Wallet` interface is exported. Is there some reasoning why the btc `Wallet` interface should not be exported?

Also, It looks like this copies code from btc `spv_wrapper` twice. dex gets away with writing this once is that possible?

JoeGruffins (approved):
Testing well.

ukane-philemon:
> I really think these changes should be done to the decred dex repo if needed. There is an obvious difference between the dcr and btc assets where the dcr `Wallet` interface is exported. Is there some reasoning why the btc `Wallet` interface should not be exported?
> 
> Also, It looks like this copies code from btc `spv_wrapper` twice. dex gets away with writing this once is that possible?

Exporting btc `Wallet` was done here: https://github.com/ukane-philemon/dcrdex/tree/btc-node.

I'll be submitting a PR upstream.

PR #421, opened 2024-01-19
Feature: Add DEX seed backup and restore

ukane-philemon:
This PR adds DEX seed export and restoration capabilities to Cryptopower.

Restoring an exported DEX with active bond:

https://github.com/crypto-power/cryptopower/assets/57448127/f8d7ae8c-067d-4ed7-beed-ee8097eda699


ukane-philemon:
UI change in https://github.com/crypto-power/cryptopower/pull/421/commits/f563675bcfd790fe33a53cdc3b5f84ca9b07a546
Screenshot 2024-01-19 at 12 08 40 PM

itswisdomagain (changes requested):
Prelim review. Will review again after the rebase.

ui/page/dcrdex/market.go
```diff
@@ -287,8 +289,16 @@ func (pg *DEXMarketPage) OnNavigatedTo() {
 	pg.ParentWindow().ShowModal(dexPasswordModal)
 }
 
+func (pg *DEXMarketPage) isDEXReset() bool {
+	if !pg.AssetsManager.DEXCInitialized() || !pg.AssetsManager.DexClient().IsDEXPasswordSet() { // dexc was reset
+		pg.ParentNavigator().ClearStackAndDisplay(NewDEXOnboarding(pg.Load, ""))
```
itswisdomagain, line 294: ClearStack? 🤔

We should only want to close the current page and any dex-related pages before it.

I think just `CloseCurrentPage()` and let the previous page be displayed. If that is a dex-related page, it should also check if dex has been reset and close itself.

ui/page/settings/app_settings_page.go
```diff
@@ -267,6 +287,18 @@ func (pg *AppSettingsPage) general() layout.Widget {
 				layout.Rigid(func(gtx C) D {
 					return pg.subSectionSwitch(gtx, values.String(values.StrTxNotification), pg.transactionNotification)
 				}),
+				layout.Rigid(func(gtx C) D {
+					if !pg.AssetsManager.DEXCInitialized() || len(pg.AssetsManager.DexClient().Exchanges()) == 0 {
+						return D{}
+					}
```
itswisdomagain, line 293: Allow backup even if exchanges have not been added yet. As long as dexc has been initialized with a password.

ui/page/settings/app_settings_page.go
```diff
@@ -375,12 +407,12 @@ func (pg *AppSettingsPage) debug() layout.Widget {
 					return pg.clickableRow(gtx, viewLogRow)
 				}),
 				layout.Rigid(func(gtx C) D {
-					if pg.AssetsManager.NetType() != libutils.Testnet || !pg.AssetsManager.DexcInitialized() {
+					if pg.AssetsManager.NetType() != libutils.Testnet || !pg.AssetsManager.DEXCInitialized() || !pg.AssetsManager.DexClient().IsDEXPasswordSet() {
```
itswisdomagain, line 410: Perhaps we need a dexClient.InitializedWithPassword method.

ui/page/settings/app_settings_page.go
```diff
@@ -647,6 +680,83 @@ func (pg *AppSettingsPage) HandleUserInteractions() {
 			pg.ParentWindow().ShowModal(currentPasswordModal)
 		}
 	}
+
+	if pg.backupDEX.Clicked() {
+		// Show modal asking for dex password and then reveal the seeds.
```
itswisdomagain, line 685:
```suggestion
		// Show modal asking for dex password and then reveal the seed.
```

ui/page/settings/app_settings_page.go
```diff
@@ -647,6 +680,83 @@ func (pg *AppSettingsPage) HandleUserInteractions() {
...
+		dexPasswordModal := modal.NewCreatePasswordModal(pg.Load).
+			EnableName(false).
+			EnableConfirmPassword(false).
+			Title(values.String(values.StrDexPassword)).
+			SetPositiveButtonCallback(func(_, password string, pm *modal.CreatePasswordModal) bool {
+				var err error
+				defer func() {
+					if err != nil {
+						pm.SetError(err.Error())
+						pm.SetLoading(false)
```
itswisdomagain, line 695: Don't see pm.SetLoading(true) anywhere here.

ui/page/settings/app_settings_page.go
```diff
@@ -647,6 +680,83 @@ func (pg *AppSettingsPage) HandleUserInteractions() {
...
+				pg.showDEXSeedModal(seed)
+				return err == nil
+			})
+
+		dexPasswordModal.SetPasswordTitleVisibility(false)
+		pg.ParentWindow().ShowModal(dexPasswordModal)
+	}
+}
+
+func (pg *AppSettingsPage) handleCopyEvent(gtx C) {
```
itswisdomagain, line 719: I don't think this is called more than once, doesn't need to be a method. If anything, we may want a util method for copying text to the clipboard.

ui/page/settings/app_settings_page.go
```diff
@@ -647,6 +680,83 @@ func (pg *AppSettingsPage) HandleUserInteractions() {
...
+			pg.copyDEXSeed.Text = values.String(values.StrCopy)
+			pg.copyDEXSeed.Color = pg.Theme.Color.Primary
+			pg.ParentWindow().Reload()
+		})
+	}
+}
+
+func (pg *AppSettingsPage) showDEXSeedModal(seed dex.Bytes) {
+	defer utils.ZeroBytes(seed)
+	pg.dexSeed = stringifySeed(seed)
```
itswisdomagain, line 734: pg.dexSeed should be []byte, not string. So you can zero once the modal is closed.

ui/page/settings/app_settings_page.go
```diff
@@ -647,6 +680,83 @@ func (pg *AppSettingsPage) HandleUserInteractions() {
...
+
+	var seedChunks []string
+	subChunkRegex := regexp.MustCompile(`.{1,8}`)
+	for _, chunk := range chunks {
+		subChunks := subChunkRegex.FindAllString(chunk, -1)
+		seedChunks = append(seedChunks, strings.Join(subChunks, "  "))
+	}
+
+	return strings.Join(seedChunks, "\n")
 }
```
itswisdomagain, line 760: Is this what upstream does? Seems new.

ui/values/localizable/en.go
```diff
@@ -886,8 +886,8 @@ const EN = `
 "integratedExchange" = "Integrated Exchange"
 "seedWords" = "Seed Words"
 "allservers" = "All servers"
-"deleteDEXData" = "Delete DEX data"
-"deleteDEXDataWarning" = "This action will delete your DEX database, including information regarding pending bonds and connected DEX servers."
+"resetDEXData" = "Reset DEX data"
+"resetDEXDataWarning" = "This action will delete your DEX database, including information regarding pending bonds and connected DEX servers. Ensure you have a backup of your DEX seeds."
```
itswisdomagain, line 890:
```suggestion
"resetDEXDataWarning" = "This action will delete your DEX database, including information regarding pending bonds and connected DEX servers. Ensure you have a backup of your DEX seed."
```

ui/values/localizable/en.go
```diff
@@ -918,4 +918,12 @@ const EN = `
...
 "dexMainnetNotReady" = "DCRDEX is currently in beta testing and has been disabled on mainnet to prevent potential loss of funds. Please switch to testnet to try it out."
 "updateDEXWalletPasswordReason" = "Your %s wallet (%s) is connected to your DEX account. Please provide your DEX password to update DEX with the new wallet password."
+"backupDEXSeed" = "Backup DEX Seed"
+"dexSeed" = "DEX Seed"
+"newDex" = "New DEX"
+"restoreDEX" = "Restore DEX"
+"restoreDEXMsg" = "Provide your DEX restoration seed to restore your DEX credentials."
+"restorationSeed" = "Restoration Seed"
+"dexResetSuccessful" = "DEX reset was successful"
+"reset" = "Reset"
```
itswisdomagain, line 928: These should be close to `resetDEXData` and `resetDEXDataWarning`.

ui/values/strings.go
```diff
@@ -215,7 +215,6 @@ const (
 	StrComplete                        = "complete"
 	StrConfirm                         = "confirm"
 	StrConfirmations                   = "confirmations"
-	StrConfirmDexReset                 = "confirmDexReset"
```
itswisdomagain, line 218: I don't see this deleted from `ui/values/localizable/en.go`.

ui/values/strings.go
```diff
@@ -1027,4 +1026,12 @@ const (
...
 	StrDexMainnetNotReady              = "dexMainnetNotReady"
 	StrUpdateDEXWalletPasswordReason   = "updateDEXWalletPasswordReason"
+	StrBackupDEXSeed                   = "backupDEXSeed"
+	StrDEXSeed                         = "dexSeed"
+	StrNewDEX                          = "newDex"
+	StrRestoreDEX                      = "restoreDEX"
+	StrRestoreDEXMsg                   = "restoreDEXMsg"
+	StrRestorationSeed                 = "restorationSeed"
+	StrDEXResetSuccessful              = "dexResetSuccessful"
+	StrReset                           = "reset"
```
itswisdomagain, line 1036: Move close to the other reset entries.

itswisdomagain (changes requested):

ui/page/settings/app_settings_page.go
ukane-philemon: Yh, added 6months ago just for better display: https://vscode.dev/github/decred/dcrdex/blob/expiredbonds/client/webserver/site/src/js/init.ts#L309

ui/page/settings/app_settings_page.go
ukane-philemon: Not called more than once but it's cleaner this way.

ui/page/settings/app_settings_page.go
ukane-philemon: It's set by the password modal before calling the callback: https://vscode.dev/github/crypto-power/cryptopower/blob/backup-dex/ui/modal/create_password_modal.go#L238

dexc/core.go
```diff
@@ -47,8 +47,12 @@ func (dc *DEXClient) InitWithPassword(pw []byte, seed []byte) error {
...
 }
 
-func (dc *DEXClient) IsDEXPasswordSet() bool {
-	return dc.IsInitialized()
+func (dc *DEXClient) IsInitialized() bool {
+	return dc != nil && dc.Core != nil
+}
+
+func (dc *DEXClient) InitializedWithPassword() bool {
+	return dc.Core.IsInitialized()
```
itswisdomagain, line 55:
```suggestion
	return dc. IsInitialized() && dc.Core.IsInitialized()
```

itswisdomagain: Also, there was a `IsDEXPasswordSet` method. Removed?

libwallet/assets_manager.go
```diff
@@ -365,7 +366,7 @@ func (mgr *AssetsManager) Shutdown() {
 	}
 
 	// Shutdown dexc before closing wallets.
-	if mgr.dexc != nil && mgr.dexc.IsInitialized() {
+	if mgr.dexc != nil && mgr.DEXCInitialized() {
```
itswisdomagain, line 369: Do you need to lock `dexcMtx` here?

libwallet/assets_manager.go
```diff
@@ -923,12 +924,13 @@ func (mgr *AssetsManager) InitializeDEX(ctx context.Context) {
 		return
 	}
 
-	if mgr.DexcInitialized() || mgr.startingDEX.Load() {
+	// Prevent multiple initialization.
+	if mgr.DEXCInitialized() || mgr.startingDEX.Load() {
```
itswisdomagain, line 928: Use `CompareAndSwap(false, true)` instead of `Load` to immediately block multiple initialization.

libwallet/assets_manager.go
```diff
@@ -979,7 +981,10 @@ func (mgr *AssetsManager) DeleteDEXData() error {
 	// both will get the ntfn?
 	<-shutdownChan // wait for shutdown
 
-	// TODO: Set mgr.dexc to nil and unregister the custom wallet constructors!
+	mgr.dexcMtx.Lock()
+	mgr.dexc = nil
+	// TODO: Also unregister the custom wallet constructors!
```
itswisdomagain, line 986: Let's take off this TODO now. The wallet constructors are now safely initialized just once for the lifetime of the app and no longer need to be unregistered.

ui/page/dcrdex/dcrdex_page.go
```diff
@@ -53,7 +53,7 @@ func NewDEXPage(l *load.Load) *DEXPage {
 		materialLoader:     material.Loader(l.Theme.Base),
 	}
 
-	if dp.AssetsManager.DexcInitialized() && dp.AssetsManager.DexClient().IsDEXPasswordSet() {
+	if dp.AssetsManager.DEXCInitialized() && dp.AssetsManager.DexClient().InitializedWithPassword() {
```
itswisdomagain, line 56: Should just call `InitializedWithPassword` only.

ui/page/dcrdex/dcrdex_page.go
```diff
@@ -123,7 +125,7 @@ func (pg *DEXPage) Layout(gtx C) D {
 		msg = values.String(values.StrDexMainnetNotReady)
 	} else if !hasMultipleWallets {
 		msg = values.String(values.StrMultipleAssetRequiredMsg)
-	} else if !pg.AssetsManager.DexcInitialized() {
+	} else if !pg.AssetsManager.DEXCInitialized() || pg.CurrentPage() == nil {
```
itswisdomagain, line 128: Should this page not close if `pg.CurrentPage()` is nil. If so, `Layout()` should just `return D{}` early on and not bother about anything else.

ui/page/dcrdex/dex_onboarding_page.go
```diff
@@ -201,7 +205,7 @@ func NewDEXOnboarding(l *load.Load, existingDEXServer string) *DEXOnboarding {
 
 	pg.isLoading = true
 	go func() {
-		<-pg.dexc.Ready()
+		<-dexc.Ready()
 		pg.isLoading = false
```
itswisdomagain, line 209: This goroutine should be in OnNavigatedTo(). That the page was intialized doesn't mean it will definitely be displayed.

Also, the display should be refreshed after toggling pg.isLoading.

ui/page/dcrdex/dex_onboarding_page.go
```diff
@@ -867,25 +886,56 @@ func (pg *DEXOnboarding) HandleUserInteractions() {
...
 			if !ok {
 				return
 			}
 
-			pg.setAddServerStep()
+			if pg.seedEditor.Editor.Text() != "" {
+				pg.dexPass = []byte(pg.passwordEditor.Editor.Text())
+
+				// Validate seed and initialize dex client.
+				seed, ok := pg.validSeed()
```
itswisdomagain, line 913: Do you mean `pg.validateSeed()`?

ui/page/dcrdex/dex_onboarding_page.go
```diff
@@ -1009,12 +1059,21 @@ func (pg *DEXOnboarding) setAddServerStep() {
...
+	dexClient := pg.AssetsManager.DexClient()
 	var xc *core.Exchange
 	var err error
 	if pg.existingDEXServer != "" { // Already registered just want to post bond.
-		xc, err = pg.dexc.Exchange(pg.bondServer.url)
+		xc, err = dexClient.Exchange(pg.bondServer.url)
+	} else if dexClient.InitializedWithPassword() {
+		var paid bool
+		xc, paid, err = dexClient.DiscoverAccount(pg.bondServer.url, pg.dexPass, pg.bondServer.cert)
+		canTrade := paid && xc.ConnectionStatus == comms.Connected && (xc.Auth.EffectiveTier > 0 || len(xc.Auth.PendingBonds) > 0)
```
itswisdomagain, line 1070: Why do we care about the connection status? Do we even care about paid? Shouldn't we only be bothered about EffectiveTier? If there are pending bonds, we want to display the bond confirmation progress?

ui/page/dcrdex/dex_onboarding_page.go
```diff
@@ -1304,13 +1365,38 @@ func (pg *DEXOnboarding) passwordsMatch(editors ...*widget.Editor) bool {
 }
 
 func (pg *DEXOnboarding) validPasswordInputs() bool {
-	validPassword := utils.EditorsNotEmpty(pg.confirmPasswordEditor.Editor)
-	if !validPassword {
+	pass := strings.TrimSpace(pg.confirmPasswordEditor.Editor.Text())
```
itswisdomagain, line 1368: Why `TrimSpace`?? Spaces could be part or a password.

ui/page/dcrdex/market.go
```diff
@@ -210,19 +207,30 @@ func NewDEXMarketPage(l *load.Load, selectServer string) *DEXMarketPage {
...
+	dexc := pg.AssetsManager.DexClient()
+	noteFeed := dexc.NotificationFeed()
 	go func() {
 		defer func() {
 			noteFeed.ReturnFeed()
 		}()
 		for {
+			// Always check if the dex client is ready. We want to exit if there
+			// was a reset.
+			if !pg.AssetsManager.DEXCInitialized() {
```
itswisdomagain, line 225: Is this the check we should be making earlier instead of `pg.isDEXReset()`? Or should we be using `pg.isDEXReset()` here and below?

itswisdomagain (changes requested):

ui/page/dcrdex/market.go
ukane-philemon: `pg.isDEXReset()` triggers and action that we only want to execute once but I can find a way to reuse the method here.

dexc/core.go
ukane-philemon: > Also, there was a IsDEXPasswordSet method. Removed?

Yes, it's the same thing.

libwallet/assets_manager.go
ukane-philemon: Oh, yh.

ui/page/dcrdex/dcrdex_page.go
ukane-philemon: > Should just call InitializedWithPassword only.

Errm the first check enures that `mgr.dexc` is non-nil before we can safely access `DEXClient`

ui/page/dcrdex/dcrdex_page.go
ukane-philemon: > Should this page not close if pg.CurrentPage() is nil.

No. We still want to display at least a message, I think.

ui/page/dcrdex/dex_onboarding_page.go
ukane-philemon: Ah, right.

ui/page/dcrdex/dcrdex_page.go
```diff
@@ -180,9 +182,9 @@ func (pg *DEXPage) OnNavigatedFrom() {}
 
 // pendingBondConfirmation is a convenience function based on arbitrary
 // heuristics to determine when to show bond confirmation step.
-func pendingBondConfirmation(am *libwallet.AssetsManager) (string, *core.BondAsset, *core.PendingBondState) {
+func pendingBondConfirmation(am *libwallet.AssetsManager, host string) (string, *core.BondAsset, *core.PendingBondState) {
 	for _, xc := range am.DexClient().Exchanges() {
-		if len(xc.Auth.PendingBonds) == 0 {
+		if host != "" && xc.Host != host || len(xc.Auth.PendingBonds) == 0 {
```
itswisdomagain, line 187: This needs parentheses.

libwallet/assets_manager.go
```diff
@@ -940,46 +943,55 @@ func (mgr *AssetsManager) InitializeDEX(ctx context.Context) {
...
 }
 
 func (mgr *AssetsManager) DeleteDEXData() error {
+	if !mgr.DEXCInitialized() {
+		return nil // nothing to do.
+	}
+
+	mgr.dexcMtx.RLock()
+	dexc := mgr.dexc
+	mgr.dexcMtx.RUnlock()
```
itswisdomagain, line 971: I think hold a write-lock till the shutdown completes so no other process gains access to a dexc instance that is about to be shutdown.

ui/page/dcrdex/market.go
```diff
@@ -506,6 +528,10 @@ func (pg *DEXMarketPage) OnNavigatedFrom() {
 // to be eventually drawn on screen.
 // Part of the load.Page interface.
 func (pg *DEXMarketPage) Layout(gtx C) D {
+	if pg.isDEXReset(true) {
+		return D{}
+	}
```
itswisdomagain, line 533: If this is the only place you pass true, then you can bring the CloseCurrentPage code here and you don't need to pass any param to the method.

ui/page/settings/app_settings_page.go
```diff
@@ -114,6 +133,7 @@ func (pg *AppSettingsPage) OnNavigatedTo() {
 // to be eventually drawn on screen.
 // Part of the load.Page interface.
 func (pg *AppSettingsPage) Layout(gtx C) D {
+	pg.handleCopyEvent(gtx)
```
itswisdomagain, line 136: handleDEXSeedCopyEvent.

ui/page/settings/app_settings_page.go
```diff
@@ -673,6 +706,81 @@ func (pg *AppSettingsPage) HandleUserInteractions() {
...
+			SetPositiveButtonCallback(func(_, password string, pm *modal.CreatePasswordModal) bool {
+				dexc := pg.AssetsManager.DexClient()
+				err := dexc.Login([]byte(password))
+				if err != nil {
+					pm.SetError(err.Error())
+					pm.SetLoading(false)
+					return false
+				}
+
+				pg.dexSeed, err = dexc.ExportSeed([]byte(password))
```
itswisdomagain, line 725: Do you need to login first? ExportSeed also takes the password.

ui/page/settings/app_settings_page.go
```diff
@@ -673,6 +706,81 @@ func (pg *AppSettingsPage) HandleUserInteractions() {
...
+				if err != nil {
+					pm.SetError(err.Error())
+					pm.SetLoading(false)
+					return false
+				}
+
+				pg.showDEXSeedModal()
+				return err == nil
```
itswisdomagain, line 733: err is nil at this point, just return true

ui/values/strings.go
```diff
@@ -1032,6 +1033,13 @@ const (
 	StrWalletsNeedToSyncMsg            = "walletsNeedToSyncMsg"
 	StrIWillSyncLater                  = "iWillSyncLater"
 	StrOkaySync                        = "okaySync"
+	StrBackupDEXSeed                   = "backupDEXSeed"
+	StrDEXSeed                         = "dexSeed"
+	StrNewDEX                          = "newDex"
+	StrRestoreDEX                      = "restoreDEX"
+	StrRestoreDEXMsg                   = "restoreDEXMsg"
+	StrRestorationSeed                 = "restorationSeed"
+	StrOptionalRestorationSeed         = "optionalRestorationSeed"
```
itswisdomagain, line 1042: I only see `backupDEXSeed`, `dexSeed` and `optionalRestorationSeed` in the en.go file.

itswisdomagain (changes requested):

ui/values/strings.go
ukane-philemon: updated

libwallet/assets_manager.go
ukane-philemon: ok.

ui/page/root/home_page.go
```diff
@@ -150,9 +151,10 @@ func (hp *HomePage) OnNavigatedTo() {
 	hp.ctx, hp.ctxCancel = context.WithCancel(context.TODO())
 
 	hp.initPageItems()
+	hp.initDEX()
 
 	go hp.CalculateAssetsUSDBalance()
-	if !hp.AssetsManager.DexcInitialized() {
+	if !hp.AssetsManager.DEXCInitialized() {
 		go hp.AssetsManager.InitializeDEX(hp.dexCtx)
```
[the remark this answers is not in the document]
ukane-philemon: Dang! Thanks!

ui/page/settings/app_settings_page.go
```diff
@@ -352,6 +341,39 @@ func (pg *AppSettingsPage) networkSettings() layout.Widget {
...
+
+					backupDEX := row{
+						title:     values.String(values.StrBackupDEXSeed),
+						clickable: pg.backupDEX,
+						label:     pg.Theme.Body2(""),
+					}
+					return pg.clickableRow(gtx, backupDEX)
+				}),
+				layout.Rigid(func(gtx C) D {
+					if pg.AssetsManager.NetType() != libutils.Testnet || !pg.AssetsManager.DEXCInitialized() || !pg.AssetsManager.DexClient().InitializedWithPassword() {
```
itswisdomagain, line 361: Okay to reset on simnet. Only avoid mainnet.

ui/values/strings.go
```diff
@@ -288,7 +288,7 @@ const (
 	StrDestinationMissing              = "destinationMissing"
 	StrDestinationModalInfo            = "destinationModalInfo"
 	StrDestinationWalletNotSynced      = "destinationWalletNotSynced"
-	StrDex                             = "dex"
+	StrDEX                             = "dex"
```
itswisdomagain, line 291: Was this not used before? Don't see previous reference being updated.

itswisdomagain (changes requested):

ui/values/strings.go
ukane-philemon: > Was this not used before?

Yh, looks like it.

ui/page/settings/app_settings_page.go
```diff
@@ -343,6 +343,10 @@ func (pg *AppSettingsPage) networkSettings() layout.Widget {
 
 func (pg *AppSettingsPage) dexSettings() layout.Widget {
 	return func(gtx C) D {
+		if pg.AssetsManager.NetType() == libutils.Mainnet {
+			return D{}
+		}
```
itswisdomagain, line 348: Won't this show a "DEX" section with nothing in it?

I think we'd get a similar display for testnet if dex isn't yet initialized with a password.

ukane-philemon: > Won't this show a "DEX" section with nothing in it? 

No, but you are right  about this "I think we'd get a similar display for testnet if dex isn't yet initialized with a password.". The checks should come early.

ukane-philemon: Updated

itswisdomagain (approved):

PR #423, opened 2024-01-20
Display executed orders with active matches as open orders

ukane-philemon:

### DIFF
--- a/ui/page/dcrdex/market.go
+++ b/ui/page/dcrdex/market.go
@@ -381,7 +381,7 @@ func (pg *DEXMarketPage) fetchOrderBook() {
 	}
 
 	// Update order form editors.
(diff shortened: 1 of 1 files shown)


ui/page/dcrdex/market.go
```diff
@@ -1801,6 +1807,15 @@ func (pg *DEXMarketPage) refreshOrders() {
 	}
 }
 
+func isActiveOrder(matches []*core.Match) bool {
```
itswisdomagain, line 1810: Pass in the `order` so this fn name makes sense. Or rename to `anyMatchActive`.

itswisdomagain (changes requested):

itswisdomagain (approved):

PR #427, opened 2024-01-20
don't init dex on mainnet and mobile, hide dex tab on mobile

itswisdomagain:


config.go
```diff
@@ -100,7 +101,7 @@ func loadConfig() (*config, error) {
 
 	// Default config
 	defaultHomeDir := dcrutil.AppDataDir("cryptopower", false)
-	if isAndroidOS {
+	if appos.Current().IsAndroid() {
```
dreacot, line 104:
```suggestion
	if appos.Current().IsMobile() {
```

dreacot (approved):

PR #435, opened 2024-01-26
Unexport `modal.SetLoading` and allow modals show loading btn

ukane-philemon:
Issue: Some modals like `InfoModal` , `TextInputModal `, and `CreateWatchOnlyModal` do not show a loading btn when the `positiveButtonClicked` callback is executed. Also it didn't fit the pattern  to have callers `m.SetLoading(false)` when they encounter an error in  the `positiveButtonClicked` callback  when we can do that once the callback returns if need be(we already do `m.SetLoading(true)` before calling the callback).

### DIFF
--- a/ui/modal/create_password_modal.go
+++ b/ui/modal/create_password_modal.go
@@ -154,7 +154,7 @@ func (cm *CreatePasswordModal) SetNegativeButtonCallback(callback func()) *Creat
 	return cm
 }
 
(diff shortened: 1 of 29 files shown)


dreacot:
please rebase

ukane-philemon:
@dreacot, I've just rebased.

dreacot (approved):

PR #438, opened 2024-01-29
dcr: Limit methods accessible on dcr.Asset using WalletHelper interface for dcr dex wallet

ukane-philemon:


dreacot (approved):

libwallet/assets/dcr/dex_wallet.go
```diff
@@ -33,17 +33,29 @@ import (
...
 }
 
+type WalletHelper interface {
+	AccountName(accountNumber int32) (string, error)
+	IsAccountMixerActive() bool
+	UnmixedAccountNumber() int32
+	MixedAccountNumber() int32
+	LockWallet()
+	UnlockWallet(pass string) error
+}
```
itswisdomagain, line 49: *dcrwallet.Wallet has methods for `AccountName`, `LockWallet` and `UnlockWallet`. Why are we using a different implementation's?

libwallet/assets/dcr/dex_wallet.go
```diff
@@ -504,21 +515,21 @@ func (dw *DEXWallet) MatchAnyScript(ctx context.Context, blockHash *chainhash.Ha
...
 func (dw *DEXWallet) AccountUnlocked(_ context.Context, _ string) (bool, error) {
-	return !dw.Wallet.IsLocked(), nil
+	return !dw.w.Locked(), nil
 }
 
 // LockAccount locks the specified account.
 // Part of the Wallet interface.
 func (dw *DEXWallet) LockAccount(_ context.Context, _ string) error {
-	dw.Wallet.LockWallet()
+	dw.helper.LockWallet()
```
itswisdomagain, line 524: If there's a reason for using the helper method rather than the wallet method, please leave a comment here.

libwallet/assets/dcr/dex_wallet.go
```diff
@@ -504,21 +515,21 @@ func (dw *DEXWallet) MatchAnyScript(ctx context.Context, blockHash *chainhash.Ha
...
 	return nil
 }
 
 // UnlockAccount unlocks the specified account or the wallet if account is not
 // encrypted.
 // Part of the Wallet interface.
 func (dw *DEXWallet) UnlockAccount(_ context.Context, pass []byte, _ string) error {
-	return dw.Wallet.UnlockWallet(string(pass))
+	return dw.helper.UnlockWallet(string(pass))
```
itswisdomagain, line 532: Same comment as above.

itswisdomagain (changes requested):

libwallet/assets/dcr/dex_wallet.go
ukane-philemon: Uhmm, not really. The previous implementation required using `Wallet.LockWallet` (not wallet.Wallet) due to some checks in this method(and others) that we didn't want to duplicate everywhere.

But maybe we can do without them now that we are using `wallet.Wallet`.

The same applies to other methods in the `WalletHelper` interface

libwallet/assets/dcr/dex_wallet.go
ukane-philemon: https://github.com/crypto-power/cryptopower/pull/438#discussion_r1474489703

libwallet/assets/dcr/dex_wallet.go
```diff
@@ -69,26 +78,26 @@ func (dw *DEXWallet) SpvMode() bool {
 // Accounts returns the names of the accounts for use by the exchange wallet.
 func (dw *DEXWallet) Accounts() dexdcr.XCWalletAccounts {
 	var accts dexdcr.XCWalletAccounts
-	accountName, err := dw.AccountName(dw.tradingAccountNumber)
+	accountName, err := dw.w.AccountName(context.Background(), uint32(dw.tradingAccountNumber))
```
ukane-philemon, line 81: This or to add `AccountName(acctNumber)` to walletHelper.

itswisdomagain: This is fine. We can also save the ctx from `dw.Connect` and use it here and anywhere else.

itswisdomagain (approved):

PR #444, opened 2024-02-01
hotfix: Fix LTC `*neutrino.ChainService.Peers()` issue

ukane-philemon:
This PR fixes an issue with LTC `*neutrino.ChainService.Peers()`. The `.Peers()` method may stall, especially if the wallet hasn't started syncing yet. We had a similar fix for BTC but the issue was not noticed earlier for LTC wallets.

EDIT: Other hotfix commits have been moved to https://github.com/crypto-power/cryptopower/pull/449 so this PR can progress quickly.

### DIFF
--- a/libwallet/assets/ltc/dex-wallet.go
+++ b/libwallet/assets/ltc/dex-wallet.go
@@ -26,6 +26,7 @@ import (
 	"github.com/btcsuite/btcwallet/waddrmgr"
 	btcwallet "github.com/btcsuite/btcwallet/wallet"
 	"github.com/btcsuite/btcwallet/wtxmgr"
+	neutrino "github.com/dcrlabs/neutrino-ltc"
 	"github.com/dcrlabs/neutrino-ltc/chain"
 	btcneutrino "github.com/lightninglabs/neutrino"
 	"github.com/lightninglabs/neutrino/headerfs"
@@ -555,12 +556,26 @@ func (s *ltcChainService) BestBlock() (*headerfs.BlockStamp, error) {
 }
 
 func (s *ltcChainService) Peers() []dexbtc.SPVPeer {
-	rawPeers := s.CS.Peers()
-	peers := make([]dexbtc.SPVPeer, len(rawPeers))
-	for i, p := range rawPeers {
-		peers[i] = p
+	// *neutrino.ChainService.Peers() may stall, especially if the wallet hasn't
+	// started sync yet. Call the method in a goroutine and wait below to see if
+	// we get a response. Return an empty slice if we don't get a response after
+	// waiting briefly.
+	rawPeersChan := make(chan []*neutrino.ServerPeer)
+	go func() {
+		rawPeersChan <- s.CS.Peers()
+	}()
+
+	select {
+	case rawPeers := <-rawPeersChan:
+		peers := make([]dexbtc.SPVPeer, 0, len(rawPeers))
+		for _, p := range rawPeers {
+			peers = append(peers, p)
+		}
+		return peers
+
+	case <-time.After(2 * time.Second):
+		return nil // CS.Peers() is taking too long to respond
 	}
-	return peers
 }
 
 func (s *ltcChainService) AddPeer(addr string) error {


ui/page/components/wallet_sync_info.go
```diff
@@ -462,6 +462,8 @@ func (wsi *WalletSyncInfo) progressStatusDetails() (int, string) {
 func (wsi *WalletSyncInfo) layoutAutoSyncSection(gtx C) D {
 	return layout.Flex{}.Layout(gtx,
 		layout.Rigid(func(gtx C) D {
+			autoSync := wsi.wallet.ReadBoolConfigValueForKey(sharedW.AutoSyncConfigKey, false)
+			wsi.syncSwitch.SetChecked(autoSync)
```
itswisdomagain, line 466: Why is this being done here? I think it should be done in `OnNavigatedTo()`?

ui/page/root/home_page.go
```diff
@@ -219,6 +220,79 @@ func (hp *HomePage) initDEX() {
 		dexPassEditor := hp.Theme.EditorPassword(new(widget.Editor), values.String(values.StrDexPassword))
 		dexPassEditor.Editor.SingleLine, dexPassEditor.IsRequired = true, true
 
+		showWalletToSyncModal := func() {
```
itswisdomagain, line 223:
```suggestion
		showWalletsToSyncModal := func() {
```

ui/page/root/home_page.go
```diff
@@ -219,6 +220,79 @@ func (hp *HomePage) initDEX() {
...
+					log.Error("dex wallet with ID %d is missing", walletID)
+					continue
+				}
+
+				if wallet.IsSynced() || wallet.IsSyncing() {
+					continue // ok
+				}
+
+				walletsToSync = append(walletsToSync, wallet)
+				walletsToSyncStr += "," + fmt.Sprintf("%s (%s)", wallet.GetWalletName(), wallet.GetAssetType())
```
itswisdomagain, line 262: The comma can go into the sprintf. There should be a space after the comma as well.
```suggestion
				walletsToSyncStr += fmt.Sprintf(", %s (%s)", wallet.GetWalletName(), wallet.GetAssetType())
```

ui/page/root/home_page.go
```diff
@@ -219,6 +220,79 @@ func (hp *HomePage) initDEX() {
...
+			}
+
+			if len(walletsToSync) == 0 {
+				return
+			}
+
+			walletSyncRequestModal := modal.NewCustomModal(hp.Load).
+				Title(values.String(values.StrWalletsNeedToSync)).
+				Body(values.StringF(values.StrWalletsNeedToSyncMsg, strings.Trim(walletsToSyncStr, ","))).
```
itswisdomagain, line 271: A preferred alternative to this would be to store the wallet names in a slice and use `strings.Join(namesSlice, ", ")` here. Don't call it `namesSlice`, that's just an example.

ui/page/root/home_page.go
```diff
@@ -219,6 +220,79 @@ func (hp *HomePage) initDEX() {
...
+						hp.Toast.NotifyError(values.String(values.StrNotConnected))
+						return false
+					}
+
+					for _, w := range walletsToSync {
+						err := w.SpvSync()
+						if err != nil {
+							log.Error(err)
+						} else {
+							w.SaveUserConfigValue(sharedW.AutoSyncConfigKey, true)
```
itswisdomagain, line 286: Uhm... why though? I think we should let the user be the one to enable auto sync. They may have it disabled for some reason. We only care that the wallet is syncing now, not that it auto-syncs next time.

ui/page/root/home_page.go
```diff
@@ -230,6 +304,11 @@ func (hp *HomePage) initDEX() {
 					}),
 				)
 			}).
+			SetNegativeButtonText(values.String(values.StrLoginLater)).
+			NegativeButtonStyle(hp.Theme.Color.OrangeRipple, hp.Theme.Color.Surface).
```
itswisdomagain, line 308: Use the same style for the "I'll sync later" button. This "Login later" button shouldn't look like a primary button, it should be less visible.

ui/page/root/home_page.go
```diff
@@ -219,6 +220,79 @@ func (hp *HomePage) initDEX() {
...
+						err := w.SpvSync()
+						if err != nil {
+							log.Error(err)
+						} else {
+							w.SaveUserConfigValue(sharedW.AutoSyncConfigKey, true)
+						}
+					}
+
+					hp.ParentWindow().Reload()
+					return true
```
itswisdomagain, line 291: Is the window reload necessary here? When you return true, it'll dismiss the modal which triggers a window reload.

ui/values/localizable/en.go
```diff
@@ -923,11 +923,12 @@ const EN = `
...
 "walletsNeedToSync" = "Wallet(s) need to sync"
-"walletsNeedToSyncMsg" = "Wallet(s) used by the DEX client needs to sync to settle active trade(s) or refund expired bonds."
+"walletsNeedToSyncMsg" = "The following wallet(s) used by the DEX client needs to sync to settle active trade(s) or refund expired bonds: %s"
 "iWillSyncLater" = "I'll sync later"
 "okaySync" = "Okay, sync"
 "paste" = "Paste"
 "backupDEXSeed" = "Backup DEX Seed"
 "dexSeed" = "DEX Seed"
 "optionalRestorationSeed" = "Restoration Seed (optional)"
+"loginLater" = "I'll login later"
```
itswisdomagain, line 933: Move this up, close to "i will sync later"

ui/values/strings.go
```diff
@@ -1038,6 +1038,7 @@ const (
 	StrBackupDEXSeed                   = "backupDEXSeed"
 	StrDEXSeed                         = "dexSeed"
 	StrOptionalRestorationSeed         = "optionalRestorationSeed"
+	StrLoginLater                      = "loginLater"
```
itswisdomagain, line 1041: Same comment as previous.

itswisdomagain (changes requested):

ui/page/root/home_page.go
ukane-philemon: Uhmm, I can remove I think. The sync status value is now fetched db right when the `layout` is drawn.

ui/page/components/wallet_sync_info.go
ukane-philemon: The value might have been updated in a gorutine and it won't reflect in the UI until another `OnNavigatedTo()` is called.

itswisdomagain, line 466: This appears to be the only place where users see a checkbox indicating that the wallet is syncing. This should use `wallet.IsSyncing() || wallet.IsSynced()` instead.

ui/page/root/home_page.go
itswisdomagain: I just left a comment that will let users know that the wallet is syncing without enabling auto-sync for the wallet.

itswisdomagain (changes requested):

ui/page/components/wallet_sync_info.go
ukane-philemon: Updated.

Users will have to use the sync btn next time they start the app.

ukane-philemon:
Notable UI changes: 

Screenshot 2024-02-02 at 10 11 01 PM
Screenshot 2024-02-02 at 10 12 13 PM

ui/page/dcrdex/market.go
```diff
@@ -1928,8 +1970,10 @@ func (pg *DEXMarketPage) orderWithLots() bool {
 	return !pg.switchLotsOrAmount.IsChecked()
 }
 
-func (pg *DEXMarketPage) noSupportedMarket() bool {
-	return pg.marketSelector.Selected() == values.String(values.StrNoSupportedMarket)
+// noMarket checks if the selected server is disconnected or has no supported
+// markets.
+func (pg *DEXMarketPage) noMarket() bool {
+	return pg.marketSelector.Selected() == values.String(values.StrNoSupportedMarket) || pg.marketSelector.Selected() == values.String(values.StrDEXServerDisconnected)
 }
```
itswisdomagain, line 1977: This is a weird check. Why checking the string instead of the dex client?

ui/values/localizable/en.go
```diff
@@ -909,12 +909,13 @@ const EN = `
...
 "noSupportedMarketMsg" = "Oops, there are no supported market at the moment. Please choose another server."
 "postBondMsg" = "You have zero effective tiers. Please post bonds to trade."
-"bondPostingInProgressMsg" = "You have zero effective tiers but bond posting is in progress. Please ensure your %s(%s) wallet account %s is funded with at least %s and is syncing."
+"bondPostingInProgressMsg" = "You have zero effective tiers but bond posting is in progress. Please ensure you have more than %s in the %q account of your %s (%s) wallet."
 "maxDEX" = "Max %s: %s"
 "availableBalance" = "Available Balance: %s"
 "missingDEXWalletMsg" = "No %s trade wallet found. Please add %s wallet to the DEX client to trade." 
 "swapAndRedeemFee" = "swap - %s, redeem - %s"
-"loginWithDEXPassword" = "Login with your DEX password."
+"loginWithDEXPassword" = "Login with your DEX password to trade on DCRDEX."
```
itswisdomagain, line 917: Isn't this message also used when asking the user to login to dex to refund expired bonds?

ui/values/localizable/en.go
```diff
@@ -909,12 +909,13 @@ const EN = `
...
+"loginDEXForPendingBonds" = "You have pending bond transaction(s). Login now else your bond transactions will wait until you do!"
```
itswisdomagain, line 918: We require login for pending bonds now?

itswisdomagain (changes requested):

ui/values/localizable/en.go
ukane-philemon: No.

ui/values/localizable/en.go
ukane-philemon: Yh, if the user cancelled login on app start.

ui/page/dcrdex/market.go
ukane-philemon: This is what the user sees atm, and I think it's safe to base our judgment on that. If there xc has reconnected, the page is instantly updated.

itswisdomagain (approved):

PR #446, opened 2024-02-01
fix crash when pasting an empty text into the editor

### LINKED ISSUE #445: app crashes when trying to paste an empty text into the editor

### LINKED ISSUE #458: Watch only balance doesn’t show on the main page either.
Watch only balance doesn’t show on the main page either.

This is all on mobile.

dreacot:
closes #445 
closes #458

- fix crash when pasting an empty text into the editor
- add watch only wallet sync card to overview page
- fix hover color for setting options on start page
- fix watch only wallet balance not being accounted for
- show wallet name in the editor while trying to rename wallet
- fix a bug where address discovery was not marked as completed for watch only wallets

### DIFF
--- a/libwallet/assets/dcr/syncnotification.go
+++ b/libwallet/assets/dcr/syncnotification.go
@@ -403,6 +403,11 @@ func (asset *Asset) discoverAddressesFinished() {
 		return
 	}
 
+	err := asset.MarkWalletAsDiscoveredAccounts() // Mark address discovery as completed.
+	if err != nil {
+		log.Error(err)
+	}
+
 	asset.stopUpdatingAddressDiscoveryProgress()
 }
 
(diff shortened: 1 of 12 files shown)


JustinBeBoy (changes requested):
Great work!

I think watch only wallet sync card should have another Icon for it
image

JustinBeBoy:
> fix watch only wallet balance not being accounted for

Why we need it? May be watch only wallet is not of them. Additional, if you want to add watch only wallet balance, it should is option and can setup on app setting

dreacot:
> Why we need it?

users are asking for it.

> can setup on app setting

we can do this in a different pr, or if users request for it

JustinBeBoy (approved):

PR #449, opened 2024-02-05
Dex Improvements

ukane-philemon:
1. Added a btn to close the required dex pass modal for active orders and/or expired bonds.
Screenshot 2024-02-01 at 2 05 39 PM

2. Improved message for dex wallet(s) that needs to turn on sync.
Screenshot 2024-02-01 at 3 10 04 AM

3. Update sync status in the database for auto-synced dex wallets.

4. Prevent interaction on the order form if the user is not logged in.
Screenshot 2024-02-02 at 10 11 01 PM

5. Request dex password on the pending bond page if the user did not log in.
Screenshot 2024-02-02 at 10 12 13 PM


ui/page/dcrdex/market.go
```diff
@@ -796,130 +828,136 @@ func (pg *DEXMarketPage) orderForm(gtx C) D {
...
 					},
 					nil, 180, pg.Theme.Color.Gray3, actionBtn)
-			}),
-		)
+			}
+
+			gtx = gtx.Disabled()
+		}
+
+		return layout.Stack{}.Layout(gtx, layout.Expanded(formLayout), layout.Stacked(overlay))
```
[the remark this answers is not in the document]
ukane-philemon: > formLayout uses a layout.Stack{Alignment: layout.NW} and is here being wrapped in a layout.Stack{}. Looks off to me.

Not clear. I don't think `overlay` should be part of `formLayout`.

ui/page/dcrdex/dex_onboarding_page.go
```diff
@@ -1282,29 +1282,39 @@ func (pg *DEXOnboarding) checkForPendingBondPayment(host string) {
...
 	}
-	pg.bondServer.url = xcHost
-	pg.bondSourceAccountSelector = components.NewWalletAndAccountSelector(pg.Load, bondAssetType)
-	ok := pg.bondSourceAccountSelector.SetSelectedAsset(bondAssetType)
-	if !ok { // impossible but can happen if user deletes wallet shortly after posting bonds.
-		pg.notifyError(values.String(values.StrNoWalletLoaded))
+
+	dexClient := pg.AssetsManager.DexClient()
+	if !dexClient.IsLoggedIn() {
+		pg.ParentWindow().ShowModal(dexLoginModal(pg.Load, dexClient, true, waitForBondTx))
```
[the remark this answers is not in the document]
ukane-philemon: > If the user dismisses this modal, he's stuck sort of. He can navigate to other parts of the app but won't be able to use the dex.

If the navigate away and back, they'll get the prompt again. Just that the confirmation page currently does not accommodate another btn suggesting to them to login.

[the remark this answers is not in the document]
ukane-philemon: > This can also allow you set a NegativeButtonCallback to show a "you need to login to confirm pending bonds" message and a login button.

I think we can skip this follow up message. They obviously saw the first message and will get another opportunity if the revisit the page.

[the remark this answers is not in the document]
itswisdomagain: > If they navigate away and back

Yes, but wouldn't be obvious. Plus the page would seem to suggest that. the app is waiting for confirmations on the pending bond, but it really isn't.



> the confirmation page currently does not accommodate another btn

Is there a button there currently?

ui/page/dcrdex/market.go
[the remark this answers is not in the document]
itswisdomagain: You currently do => Stack { formLayout, overlay }

I'm saying, if there's no overlay, just return formLayout only.

ui/page/dcrdex/dex_onboarding_page.go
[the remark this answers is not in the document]
ukane-philemon: Updated. We now show a login form if the user cancels the modal.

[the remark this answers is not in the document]
ukane-philemon: Yh, at the end of the page only if the user has an effective tier and pending bond too.

ui/page/dcrdex/dex_onboarding_page.go
```diff
@@ -984,58 +1014,103 @@ func (pg *DEXOnboarding) HandleUserInteractions() {
...
 			pg.isLoading = true
-			if !dexc.InitializedWithPassword() {
-				go func() {
-					// Set password.
-					pg.dexPass = []byte(pg.passwordEditor.Editor.Text())
-					if err := dexc.InitWithPassword(pg.dexPass, nil); err != nil {
+			if dexc.InitializedWithPassword() {
+				if !dexc.IsLoggedIn() {
+					pg.showDEXPasswordModal(pg.postBond)
+				} else {
```
itswisdomagain, line 1021: It shouldn't even get to the point of attempting to post a bond if the user isn't logged in.

ui/page/dcrdex/dex_onboarding_page.go
```diff
@@ -966,10 +987,19 @@ func (pg *DEXOnboarding) HandleUserInteractions() {
...
+					pg.isLoading = false
+				}()
+				break
+			}
+
+			// dexc password is already set and the dex password is not cached.
+			// Prompt the user to login now so we can use
+			// dexClient.DiscoverAccount in
+			// pg.connectServerAndPrepareForBonding.
+			pg.showDEXPasswordModal(pg.connectServerAndPrepareForBonding)
```
itswisdomagain, line 1002: Reverse this if statement so the `go func` isn't nested.

ui/page/dcrdex/market.go
```diff
@@ -1560,18 +1604,25 @@ func (pg *DEXMarketPage) HandleUserInteractions() {
...
-			}
+			ordID := ord.ID
+			go func() {
+				err := dexc.Cancel(ordID)
+				if err != nil {
+					pg.notifyError(fmt.Errorf("Error canceling order: %v", err).Error())
+				} else {
+					pg.ParentWindow().Reload()
+				}
+			}()
```
itswisdomagain, line 1625:
```suggestion
			go func(ordID string) {
				err := dexc.Cancel(ordID)
				if err != nil {
					pg.notifyError(fmt.Errorf("Error canceling order: %v", err).Error())
				} else {
					pg.ParentWindow().Reload()
				}
			}(ord.ID)
```

ui/page/root/home_page.go
```diff
@@ -233,6 +233,76 @@ func (hp *HomePage) initDEX() {
 		dexPassEditor := hp.Theme.EditorPassword(new(widget.Editor), values.String(values.StrDexPassword))
 		dexPassEditor.Editor.SingleLine, dexPassEditor.IsRequired = true, true
 
+		showWalletsToSyncModal := func() {
```
itswisdomagain, line 236: New thought: request wallet sync before unlocking the dex client. It'd be better for the dex client to find wallets already syncing by the time the dex is logged in.

ui/values/localizable/en.go
```diff
@@ -923,13 +924,16 @@ const EN = `
...
+"walletsNeedToSyncMsg" = "The following wallet(s) used by the DEX client needs to sync to settle active trade(s) or refund expired bonds: %s"
 "iWillSyncLater" = "I'll sync later"
+"iWillLoginLater" = "I'll login later"
 "okaySync" = "Okay, sync"
 "paste" = "Paste"
 "backupDEXSeed" = "Backup DEX Seed"
 "dexSeed" = "DEX Seed"
 "optionalRestorationSeed" = "Restoration Seed (optional)"
+"dexServerDisconnected" = "DEX server is disconnected"
+"dexMsgAfterReset" = "It seems DCRDEX has been reset, please revisit this page."
```
itswisdomagain, line 936: I'm not sure ordinary users will understand how to "revisit" a page. Why can't we reopen the page for them automatically?

itswisdomagain (changes requested):

ui/values/localizable/en.go
ukane-philemon: > Why can't we reopen the page for them automatically? 

The page needs an initialized dexc client which we only prepare in the trade page. If the last page on the stack after the user resets dex was the dex page, we might not have an initialized client by the time they close the settings page and go back to the dex page.

But, I think we can show the splash page instead until dexc is ready.

ui/page/root/home_page.go
ukane-philemon: ok

ui/page/dcrdex/dex_onboarding_page.go
ukane-philemon: Updated.

ui/page/dcrdex/dcrdex_page.go
```diff
@@ -75,21 +75,31 @@ func (pg *DEXPage) ID() string {
...
 	if pg.CurrentPage() != nil {
 		pg.CurrentPage().OnNavigatedTo()
+	} else {
+		pg.setPage()
+	}
+}
+
+// setPage starts a goroutine that waits for dexc to get ready before displaying
+// an appropriate page.
+func (pg *DEXPage) setPage() {
```
itswisdomagain, line 91: If this is a goroutine that eventually displays the current page, rename to `prepareInitialPage`.

ui/page/dcrdex/dcrdex_page.go
```diff
@@ -101,19 +111,14 @@ func (pg *DEXPage) OnNavigatedTo() {
 		}
 
 		pg.dexIsLoading = false
+		pg.showSplashPage = false
```
itswisdomagain, line 114: Move these code to before you call `pg.Display`.

ui/page/dcrdex/dcrdex_page.go
```diff
@@ -125,14 +130,23 @@ func (pg *DEXPage) Layout(gtx C) D {
...
 
 	if msg != "" {
 		return components.DisablePageWithOverlay(pg.Load, nil, gtx, msg, "", actionBtn)
 	}
 
+	if pg.showSplashPage || pg.dexIsLoading {
+		return pg.Theme.List(pg.splashPageContainer).Layout(gtx, 1, func(gtx C, i int) D {
+			return pg.splashPage(gtx)
+		})
+	}
```
itswisdomagain, line 148: I think return this to the top of this method. With this change, mainnet will never see the splash page and the user will also not see the splash page if he doesn't have multiple wallets. The splash page should display before any other page.

ui/page/dcrdex/dcrdex_page.go
```diff
@@ -125,14 +130,23 @@ func (pg *DEXPage) Layout(gtx C) D {
 		msg = values.String(values.StrDexMainnetNotReady)
 	} else if !hasMultipleWallets {
 		msg = values.String(values.StrMultipleAssetRequiredMsg)
-	} else if !pg.AssetsManager.DEXCInitialized() || pg.CurrentPage() == nil {
-		msg = values.String(values.StrDEXInitErrorMsg)
+	} else if !pg.AssetsManager.DEXCInitialized() || pg.CurrentPage() == nil { // dexc must have been reset.
+		pg.showSplashPage = true
+		if !pg.dexIsLoading {
+			pg.setPage()
+		}
```
itswisdomagain, line 137: This check should also happen at the top of this method.

ui/page/dcrdex/dex_onboarding_page.go
```diff
@@ -965,6 +986,13 @@ func (pg *DEXOnboarding) HandleUserInteractions() {
 			}
 
 			pg.bondServer = serverInfo
+			if dexc.InitializedWithPassword() && len(pg.dexPass) == 0 {
```
itswisdomagain, line 989: Is it possible to get here if dexc has not been initialized with password? If yes, how is that handled? If no, then `dexc.InitializedWithPassword()` doesn't need to be part of the if statement.

ui/page/dcrdex/dex_onboarding_page.go
```diff
@@ -984,58 +1012,99 @@ func (pg *DEXOnboarding) HandleUserInteractions() {
...
+				// Login.
+				err := dexc.Login(pg.dexPass)
+				if err != nil {
+					pg.isLoading = false
+					pg.notifyError(err.Error())
+					return
+				}
+
+				pg.postBond()
+			}()
```
itswisdomagain, line 1044: The if statement above can be placed inside this goroutine, no need for 2 goroutines. They both eventually call `pg.postBond()`, so start only one goroutine, inside it, check if you need to call dexc.Init before calling pg.postBond.

ui/page/dcrdex/dex_onboarding_page.go
```diff
@@ -1285,29 +1358,42 @@ func (pg *DEXOnboarding) checkForPendingBondPayment(host string) {
...
+	dexClient := pg.AssetsManager.DexClient()
+	if dexClient.IsLoggedIn() {
+		waitForBondTx()
 		return
 	}
 
-	pg.waitForConfirmationAndListenForBlockNotifications()
+	dexPasswordModal := dexLoginModal(pg.Load, dexClient, waitForBondTx)
+	dexPasswordModal.SetDescription(values.String(values.StrLoginDEXForPendingBonds))
+	dexPasswordModal.SetNegativeButtonCallback(waitForBondTx) // We'll display a form for them to login.
```
itswisdomagain, line 1395: Both positive and negative callbacks do the same thing?

ui/page/dcrdex/dex_onboarding_page.go
```diff
@@ -984,58 +1012,99 @@ func (pg *DEXOnboarding) HandleUserInteractions() {
...
-				})
-			dexPasswordModal.SetPasswordTitleVisibility(false)
-			pg.ParentWindow().ShowModal(dexPasswordModal)
+				pg.isLoading = false
+			}()
 		}
 	}
 }
 
+func (pg *DEXOnboarding) showDEXPasswordModal(callbackFn func()) {
```
itswisdomagain, line 1082: Can this not use the `dexLoginModal` helper fn?

itswisdomagain (changes requested):

ui/page/dcrdex/dcrdex_page.go
ukane-philemon: > With this change, mainnet will never see the splash page and the user will also not see the splash page if he doesn't have multiple wallets.

Yes.

> The splash page should display before any other page.

Okay.

ui/page/dcrdex/dex_onboarding_page.go
ukane-philemon: > Is it possible to get here if dexc has not been initialized with password? If yes, how is that handled? 

Yes, a password is required for users to create a new dex instance from seed.  If this happens and the user navigates away from the page, we need to ask for the password because it's no longer cached.

ui/page/dcrdex/dcrdex_page.go
ukane-philemon: These values are updated in a goroutine and will be seen the the other process handling Layout(). This means we might end up not having a page to display until the next Layout() call.

ui/page/dcrdex/dex_onboarding_page.go
ukane-philemon: Yes. The `waitForBondTx` prepares the `onBoardingStepWaitForConfirmation` page and starts a listener for the pending bond. IF the user logs in, will fall through to display the `onBoardingStepWaitForConfirmation`.

ui/page/dcrdex/dcrdex_page.go
```diff
@@ -101,32 +110,38 @@ func (pg *DEXPage) OnNavigatedTo() {
...
 func (pg *DEXPage) Layout(gtx C) D {
-	if pg.isDexFirstVisit || pg.dexIsLoading {
+	isMainnet := pg.AssetsManager.NetType() == utils.Mainnet
+	if !isMainnet && (!pg.AssetsManager.DEXCInitialized() || pg.CurrentPage() == nil) { // dexc must have been reset.
+		pg.showSplashPage = true
+		if !pg.dexIsLoading {
+			pg.ParentNavigator().CloseAllPages()
+			pg.prepareInitialPage()
+		}
+	}
```
itswisdomagain, line 128: Previously, `!pg.AssetsManager.DEXCInitialized() || pg.CurrentPage() == nil` was treated as dex init error. If there was actually a dex init error, this new code will just keep repeating this block of code. It would mostly be a no-op anyway, but should we show a message to the user if dex init fails?

ui/page/dcrdex/dex_onboarding_page.go
```diff
@@ -1271,6 +1328,10 @@ func (pg *DEXOnboarding) waitForConfirmationAndListenForBlockNotifications() {
 	asset.RemoveTxAndBlockNotificationListener(DEXOnboardingPageID)
 	asset.AddTxAndBlockNotificationListener(&sharedW.TxAndBlockNotificationListener{
 		OnBlockAttached: func(_ int, _ int32) {
+			if pg.AssetsManager.DEXCInitialized() && !pg.AssetsManager.DexClient().IsLoggedIn() {
+				// Don't update conf if we are not yet logged in.
+				return
+			}
 			pg.bondConfirmationInfo.currentBondConf++
 			pg.ParentWindow().Reload()
```
itswisdomagain, line 1336: If it is not initialized, confs should be updated? I think just check if it's not logged in, regardless of whether it is initialized or not. Or check `if not initialized OR not logged in`.

itswisdomagain (approved):
Minor issues that should be addressed in a follow up PR.

PR #493, opened 2024-03-29
dcr: Avoid freeze inside db transaction.

JoeGruffins:
In dcr theres a request for the current wallet inside a db transaction, so if a user closes a wallet during *asset.IndexTransactions() It will get the lock at *dcrLoader.UnloadWallet() but then get stuck at l.db.Close() because it cannot close the db while a transation is taking place. So the request in the transation gets stuck trying to get the lock. If that makes sense.

### DIFF
--- a/libwallet/internal/loader/dcr/loader.go
+++ b/libwallet/internal/loader/dcr/loader.go
@@ -50,6 +50,10 @@ type dcrLoader struct {
 	mixSplitLimit           int
 
 	mu sync.RWMutex
+
+	// dbMutex should be held when db transactions would circle back around
+	// and hold the mu lock to prevent a freeze.
+	dbMutex *sync.Mutex
 }
 
 // StakeOptions contains the various options necessary for stake mining.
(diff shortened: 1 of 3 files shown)


JoeGruffins:
I was trying to think of a better way, let me look again.

JoeGruffins:
Keeping the changes local to dcr https://github.com/crypto-power/cryptopower/compare/121f5eb6b488917cfc3eb2a6f3c431ece6f05179..352b9773e031f718f746c814abbb97cccaf20fb4

libwallet/internal/loader/dcr/loader.go
```diff
@@ -50,6 +50,10 @@ type dcrLoader struct {
 	mixSplitLimit           int
 
 	mu sync.RWMutex
+
+	// dbMutex should be held when db transactions would circle back around
+	// and hold the mu lock to prevent a freeze.
+	dbMutex *sync.Mutex
```
JustinBeBoy, line 56: can we use mu for all? I mean pass mutex from wallet.go to dcr loader

JoeGruffins: No. As the comment says below `This allows callers to call *dcrLoader methods inside of db transactions.`

PR #519, opened 2024-05-13
improve dark mode

JustinBeBoy:
This PR improve some layout on dark mode:

Before:

bug6
bug5
bgu4
bug3
bug2
bug1

After:

r5
r4
r3
r2
r1


ukane-philemon (approved):
LGTM

PR #522, opened 2024-05-16
Multi: Clean up

ukane-philemon:
- Remove unused db methods
- Refactor
- Rename some db methods

### DIFF
--- a/ui/page/exchange/create_order_page.go
+++ b/ui/page/exchange/create_order_page.go
@@ -117,6 +117,11 @@ type orderData struct {
 }
 
 func NewCreateOrderPage(l *load.Load) *CreateOrderPage {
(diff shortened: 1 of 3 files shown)


ui/page/exchange/create_order_page.go
```diff
@@ -133,7 +138,7 @@ func NewCreateOrderPage(l *load.Load) *CreateOrderPage {
 			Axis:      layout.Vertical,
 		}},
 		startTradingBtn: l.Theme.Button(values.String(values.StrStartTrading)),
-		isFirstVisit:    l.AssetsManager.ReadBoolValue(sharedW.IsCEXFirstVisitConfigKey, true),
```
ukane-philemon, line 136: This always fail before now because the last arg is not a pointer.

dreacot (approved):

PR #525, opened 2024-05-18
Support 12 and 24 word seeds from other wallet

### LINKED ISSUE #518: restore: Current 12 word logic is not compatible with trust / exodus / trezor wallets.
The current restore from seed for 12 words does not match that in some mobile wallets. The ones I know of use the trezor library.

Currently we do:
https://github.com/crypto-power/cryptopower/blob/e507ac0ff880c0578040308fc560317420174dbe/libwallet/assets/wallet/wallet_utils.go#L249

But to be compatible with the mentioned wallets, it would need to be:
```
import (
	"crypto/sha512"

	"golang.org/x/crypto/pbkdf2"
)

hashedSeed := pbkdf2.Key([]byte(seedMnemonic), []byte("mnemonic"), 2048, 64, sha512.New)
```

It's important to note one difference here is that our current logic allows going from seed <-> mnemonic back and forth but the trezor way only allows for a one way mnemoinc -> seed

One idea to handle this is another seed type that uses the trezor logic.

Also leaving some test data that can be used: https://github.com/trustwallet/wallet-core/blob/484dc2a98c7a7ed9023ebae6fe30b1e331fff6e0/trezor-crypto/crypto/tests/test_check.c#L5146

JustinBeBoy:
Close #518 

This PR support 12 and 24 word seed from other wallet, update logic when restore wallet and create wallet

**Screenshot**

seed2

The input hex need the dropdown because we need to know the entropy use BIP32 or BIP39
seed1

### DIFF
--- a/libwallet/assets/ltc/utils.go
+++ b/libwallet/assets/ltc/utils.go
@@ -62,7 +62,7 @@ func (asset *Asset) DeriveAccountXpub(seedMnemonic string, wordSeedType sharedW.
 	if wordSeedType == sharedW.WordSeed33 {
 		seed, err = walletseed.DecodeUserInput(seedMnemonic)
 	} else {
-		seed, err = bip39.EntropyFromMnemonic(seedMnemonic)
+		seed, err = bip39.NewSeedWithErrorChecking(seedMnemonic, "")
 	}
 	if err != nil {
 		return "", err
--- a/libwallet/assets/ltc/utils.go
+++ b/libwallet/assets/ltc/utils.go
@@ -3,14 +3,12 @@ package ltc
 import (
 	"encoding/binary"
 
-	"decred.org/dcrwallet/v3/walletseed"
 	sharedW "github.com/crypto-power/cryptopower/libwallet/assets/wallet"
 	"github.com/crypto-power/cryptopower/libwallet/utils"
 	"github.com/ltcsuite/ltcd/chaincfg"
 	"github.com/ltcsuite/ltcd/ltcutil"
 	"github.com/ltcsuite/ltcd/ltcutil/hdkeychain"
 	"github.com/ltcsuite/ltcwallet/waddrmgr"
-	"github.com/tyler-smith/go-bip39"
 )
 
 const (
@@ -58,12 +56,7 @@ func (asset *Asset) ToAmount(v int64) sharedW.AssetAmount {
 
 // DeriveAccountXpub derives the xpub for the given account.
 func (asset *Asset) DeriveAccountXpub(seedMnemonic string, wordSeedType sharedW.WordSeedType, account uint32, params *chaincfg.Params) (xpub string, err error) {
-	var seed []byte
-	if wordSeedType == sharedW.WordSeed33 {
-		seed, err = walletseed.DecodeUserInput(seedMnemonic)
-	} else {
-		seed, err = bip39.NewSeedWithErrorChecking(seedMnemonic, "")
-	}
+	seed, err := sharedW.DecodeSeedMnemonic(seedMnemonic, asset.Type, wordSeedType)
 	if err != nil {
 		return "", err
 	}
--- a/libwallet/assets/wallet/wallet_shared.go
+++ b/libwallet/assets/wallet/wallet_shared.go
@@ -27,7 +27,7 @@ type Wallet struct {
 	db        *storm.DB
 	logDir    string
 
-	EncryptedSeed         []byte
+	EncryptedMnemonic     []byte
 	IsBackedUp            bool
 	IsRestored            bool
 	HasDiscoveredAccounts bool
@@ -257,7 +257,7 @@ func (wallet *Wallet) IsWalletBackedUp() bool {
 func (wallet *Wallet) HasWalletSeed() bool {
 	wallet.mu.RLock()
 	defer wallet.mu.RUnlock()
-	return len(wallet.EncryptedSeed) > 0
+	return len(wallet.EncryptedMnemonic) > 0
 }
 
 func (wallet *Wallet) GetWalletID() int {
@@ -344,7 +344,7 @@ func CreateNewWallet(pass *AuthInfo, loader loader.AssetLoader,
 		return nil, err
 	}
 
-	encryptedSeed, err := encryptWalletMnemonic([]byte(pass.PrivatePass), mnemonic)
+	encryptedMnemonic, err := encryptWalletMnemonic([]byte(pass.PrivatePass), mnemonic)
 	if err != nil {
 		return nil, err
 	}
@@ -356,7 +356,7 @@ func CreateNewWallet(pass *AuthInfo, loader loader.AssetLoader,
 		rootDir:               params.RootDir,
 		logDir:                params.LogDir,
 		CreatedAt:             time.Now(),
-		EncryptedSeed:         encryptedSeed,
+		EncryptedMnemonic:     encryptedMnemonic,
 		PrivatePassphraseType: pass.PrivatePassType,
 		HasDiscoveredAccounts: true,
 		Type:                  assetType,
@@ -474,7 +474,7 @@ func RestoreWallet(seedMnemonic string, pass *AuthInfo, loader loader.AssetLoade
 ) (*Wallet, error) {
 	// Ensure the encrypted seeds are available before creating wallet so we can
 	// return early.
-	encryptedSeed, err := encryptWalletMnemonic([]byte(pass.PrivatePass), seedMnemonic)
+	EncryptedMnemonic, err := encryptWalletMnemonic([]byte(pass.PrivatePass), seedMnemonic)
 	if err != nil {
 		log.Errorf("wallet.createWallet: error encrypting wallet seed: %v", err)
 		return nil, err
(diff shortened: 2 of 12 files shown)


libwallet/assets/wallet/wallet_shared.go
```diff
@@ -339,12 +339,12 @@ func CreateNewWallet(pass *AuthInfo, loader loader.AssetLoader,
...
 	}
 
-	seed, err := generateSeed(assetType, pass.WordSeedType)
+	mnemonic, err := generateMnemonic(pass.WordSeedType)
 	if err != nil {
 		return nil, err
 	}
 
-	encryptedSeed, err := encryptWalletSeed([]byte(pass.PrivatePass), seed)
+	encryptedSeed, err := encryptWalletMnemonic([]byte(pass.PrivatePass), mnemonic)
```
ukane-philemon, line 347: `encryptedSeed` needs renaming.

libwallet/assets/wallet/wallet_utils.go
```diff
@@ -133,7 +132,7 @@ func (wallet *Wallet) VerifySeedForWallet(seedMnemonic, privpass string) (bool,
 	wallet.mu.RLock()
 	defer wallet.mu.RUnlock()
 
-	decryptedSeed, err := decryptWalletSeed([]byte(privpass), wallet.EncryptedSeed)
+	decryptedSeed, err := decryptWalletMnemonic([]byte(privpass), wallet.EncryptedSeed)
```
ukane-philemon, line 135: Same as my comment above, but can ignore for `wallet.EncryptedSeed`, cuz it'll break existing wallets.

libwallet/assets/wallet/wallet_utils.go
```diff
@@ -161,23 +160,23 @@ func naclLoadFromPass(pass []byte) (nacl.Key, error) {
...
-func decryptWalletSeed(pass []byte, encryptedSeed []byte) (string, error) {
+// decryptWalletMnemonic decrypts the encryptedMnemonic with secretbox.EasyOpen using pass.
+func decryptWalletMnemonic(pass []byte, encryptedMnemonic []byte) (string, error) {
 	key, err := naclLoadFromPass(pass)
 	if err != nil {
 		return "", err
 	}
 
-	decryptedSeed, err := secretbox.EasyOpen(encryptedSeed, key)
+	decryptedSeed, err := secretbox.EasyOpen(encryptedMnemonic, key)
```
ukane-philemon, line 179: same as my comment above

libwallet/assets/wallet/wallet_utils.go
```diff
@@ -187,51 +186,37 @@ func decryptWalletSeed(pass []byte, encryptedSeed []byte) (string, error) {
...
 	}
 
 	if len(entropy) > 0 {
 		if wordSeedType == WordSeed33 {
+			// Generate 33-word seeds from PGWord list
 			return walletseed.EncodeMnemonic(entropy), nil
 		}
 		// Create Seed phrase from entropy
-		// Use bip39 for 12-word seeds and 24-word seeds
+		// Use bip39 for generate 12-word seeds and 24-word seeds
```
ukane-philemon, line 208: nit `for generate` -> `to generate`

ukane-philemon (changes requested):
First pass, mostly nits. unTested.

JoeGruffins (commented):
This doesn't seem to work with a 12 word phrase from exodus/trust wallet. Maybe they do not have a check sum encoded? Can you try on mainnet with words from trust wallet?

Also, this is big, but my concern about making this change was that people that already have a btc or ltc wallet will no longer be able to restore from their words because the logic is different. If @dreacot is fine with this ok

libwallet/assets/ltc/utils.go
```diff
@@ -62,7 +62,7 @@ func (asset *Asset) DeriveAccountXpub(seedMnemonic string, wordSeedType sharedW.
 	if wordSeedType == sharedW.WordSeed33 {
 		seed, err = walletseed.DecodeUserInput(seedMnemonic)
 	} else {
-		seed, err = bip39.EntropyFromMnemonic(seedMnemonic)
+		seed, err = bip39.NewSeedWithErrorChecking(seedMnemonic, "")
```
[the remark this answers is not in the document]
JoeGruffins: also for `deriveBIP44AccountXPubsForDCR` in dcr

ui/page/components/seed_restore_page.go
```diff
@@ -448,7 +448,7 @@ func (pg *SeedRestore) verifySeeds() bool {
 
 	if isValid {
 		pg.seedPhrase = seedphrase
-		if !sharedW.VerifySeed(pg.seedPhrase, pg.walletType, pg.getWordSeedType()) {
+		if !sharedW.VerifyMnemonic(pg.seedPhrase, pg.walletType, pg.getWordSeedType()) {
```
JoeGruffins, line 451: It looks like there is an extra space in the pg.seedPhrase that is preventing the checksum from verifying. Need to trim that before doing anything with it. This is causing the verification to fail for me.

dreacot:
> Also, this is big, but my concern about making this change was that people that already have a btc or ltc wallet will no longer be able to restore from their words because the logic is different. 

No, this is not ideal, i'm yet to review the codebase but I discussed with @JustinBeBoy before this PR and he says restoring old wallets would not be a problem

JoeGruffins:
Panic restoring btc 12 words:
```
2024-06-05 15:10:07.733 [INF] SHWL: Created Wallet
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0x14d8fa6]

goroutine 71 [running]:
github.com/crypto-power/cryptopower/app.(*GenericPageModal).ParentNavigator(...)
        /home/joe/git/cryptopower/app/genericpagemodal.go:43
github.com/crypto-power/cryptopower/ui/page/components.(*SeedRestore).HandleUserInteractions.func1({0xc0000eca08?, 0x0?}, {0xc002ee6010, 0x3}, 0xc00027ea08)
        /home/joe/git/cryptopower/ui/page/components/seed_restore_page.go:549 +0x2c6
github.com/crypto-power/cryptopower/ui/modal.(*CreatePasswordModal).Handle.func1()
        /home/joe/git/cryptopower/ui/modal/create_password_modal.go:241 +0x6d
created by github.com/crypto-power/cryptopower/ui/modal.(*CreatePasswordModal).Handle in goroutine 58
        /home/joe/git/cryptopower/ui/modal/create_password_modal.go:240 +0x3b3
```

JoeGruffins:
That panic is also in master when restoring from 12 words. New wallet.

JoeGruffins:
The default for all new wallets is 33 words, should make that 12.

ui/page/components/restore_page.go
```diff
@@ -331,3 +370,21 @@ func (pg *Restore) restoreFromSeedEditor() {
...
+			return sharedW.WordSeed24, nil
+		case 33:
+			return sharedW.WordSeed33, nil
+		default:
+			return sharedW.NoneWordSeed, nil
+		}
+	}
+
+	return sharedW.NoneWordSeed, fmt.Errorf("invalid word seed")
+}
```
dreacot, line 390:
```suggestion
func getWordSeedTypeFromSeed(slideWords []string) (sharedW.WordSeedType, error) {
	switch len(slideWords) {
	case 12:
		return sharedW.WordSeed12, nil
	case 24:
		return sharedW.WordSeed24, nil
	case 33:
		return sharedW.WordSeed33, nil
	default:
		return sharedW.NoneWordSeed, fmt.Errorf("invalid word seed")
	}
}
```

ui/page/components/seed_restore_page.go
```diff
@@ -54,9 +54,9 @@ type SeedRestore struct {
 	optionsMenuCard cryptomaterial.Card
 	window          app.WindowNavigator
 
-	suggestions    []string
-	allSuggestions []string
-	seedMenu       []seedItemMenu
+	suggestions []string
+	// allSuggestions []string
```
dreacot, line 58: why is this commented?

ui/page/components/seed_restore_page.go
```diff
@@ -110,7 +110,7 @@ func NewSeedRestorePage(l *load.Load, walletName string, walletType libutils.Ass
 	pg.initSeedMenu()
 
 	// set suggestions
-	pg.allSuggestions = getWordSeedType().AllWords()
+	// pg.allSuggestions = getWordSeedType().AllWords()
```
dreacot, line 113: why is this commented?

ui/page/components/seed_restore_page.go
```diff
@@ -473,7 +473,7 @@ func (pg *SeedRestore) verifySeeds() bool {
 }
 
 func (pg *SeedRestore) resetSeeds() {
-	pg.allSuggestions = pg.getWordSeedType().AllWords()
+	// pg.allSuggestions = pg.getWordSeedType().AllWords()
```
dreacot, line 476: why is this commented?

libwallet/assets/wallet/wallet_utils.go
```diff
@@ -161,92 +160,89 @@ func naclLoadFromPass(pass []byte) (nacl.Key, error) {
...
-		if err != nil {
-			return "", err
-		}
-	case utils.LTCWalletAsset:
-		entropy, err = ltchdkeychain.GenerateSeed(length)
-		if err != nil {
-			return "", err
-		}
+
+	entropy, err = btchdkeychain.GenerateSeed(length)
```
dreacot, line 197: i'm guessing the reason only btchdkeychain was used is because all other assets have the same underlying logic in their keychain, in that case, a comment should b left tot hat effect

dreacot (changes requested):

dreacot (approved):

Issue #548, opened 2024-06-22
Send from Unmixed or Default Account between the same wallet seems disabled


dmigwi:
<img width="876" alt="Screenshot 2024-06-22 at 15 00 14" src="https://github.com/crypto-power/cryptopower/assets/22055953/26132afb-5464-45a5-918e-a80d1fff34f5">

dmigwi:
`SendMax` feature also seems inactive too.

PR #550, opened 2024-06-22
Fix the accounts selector during intra-accounts funds transfer

dmigwi:
closes: https://github.com/crypto-power/cryptopower/issues/548

During Intra-Accounts funds transfer (transfer between accounts of the same wallet) only the mixed account are available for selection.

**To enable the intra-accounts transfer the following conditions are required:**
1. There is only one recipient instance available.
2. Both (i.e. source and recipient accounts) are from the same wallet.
3. Source account selected must have a spendable balance
4. Recipient's wallet tab option must be active.
5. The destination and source accounts must be different.

<h2>Source Account Selector When Intra-Accounts Selector is modified based on the above conditions</h2>
Screenshot 2024-06-22 at 19 29 15

_NB: By default the info text above is always hidden until the info button is clicked._

<h2>Source Account Selector during all other funds transfer cases</h2>
Screenshot 2024-06-22 at 19 28 53

_NB: By default the info text above is always hidden until the info button is clicked._


ui/page/send/page.go
```diff
@@ -218,7 +218,21 @@ func (pg *Page) initializeAccountSelectors() {
 				// only mixed accounts can send to address/wallets for wallet with privacy setup
 				// don't need to check account the same with destination account
 				accountIsValid = account.Number == load.MixedAccountNumber(pg.selectedWallet)
+
+				// For an Intra-Accounts to happen transfer the bare minimum expected is that:
+				// 1. There is only one recipient instance available.
+				// 2. Both (i.e. source and recipient) use the same wallet.
+				// 3. Source account selected must have a spendable balance
+				// 4. Recipient's wallet tab option must be active.
```
dreacot, line 226: this no.4 isn't clear what it means

dreacot:
i noticed you are following the pattern with the info icon present, but i believe that should be a modal instead, while the info icon should be right beside the from text

ui/page/send/page.go
dmigwi: There are two tabs on the recipient info; "Address" and "Wallets". To activate the "wallets" tab, it needs to be currently on display.

dmigwi:
> i noticed you are following the pattern with the present info icon, but i believe that should be a modal instead, while the info icon should be right beside the from text

@dreacot I am not sure I follow. I am lost because I have changed the current functionality of the info button. I just added more description text to the current functionality.

dreacot (approved):

PR #602, opened 2024-07-25
Update some logic in send page, receive page

### LINKED ISSUE #592: send page bugs
- [ ] trying to do a test send from one wallet to another using 'Wallets' option on send page. Destination account provides nothing to select, so its no possible to complete the send. Appears to be an issue if the destination account has a 0 balance.
- [x] Click on the QR code Or the address to copy this text, Position of the toast is not shown in the right place
- [x] "Amount (DCR)" input show place holder when input selected
- [x] Remove hover on "MAX" button, change text to "Max"
- [ ] Keep the password confirmation modal in send page always visible even if an incorrect password is entered
- [x] 'You're about to send' page, 'To' label and address are out of bottom vertical alignment. Grey line in the middle is oddly placed and has inconsistent white space above and below.
- [x]  Clicking on send appears to show a partial circle that I believe is suppose to be an progress animation, but there is no animation.
- [ ] Cannot click the send button if the wallet does not have a password
- [x] Cannot focus on password input
- [x] Clear error when updating password input
- [x] Clicking in the field and holding down the mouse button for some reason causes the 'Spending passphrase' title to disappear.
- [x] Expected behaviour for me after transaction send is to go to the view of the transaction, rather then returning to empty send page. At minimum, have a button to click from the 'Sent!' dialog that will take the user there.
- [x] Show loading indicator instead of "No transaction" text when selecting between 'Regular Transactions' and 'Staking Transactions'
- [x] Add X or Cancel button to Send/Receive Modal to close modal

JustinBeBoy:
Close #592 

This PR update some logic follow issue check list


dreacot:
I have a wallet with just one account, that is the sending wallet, and i had selected the one account as the source account

i chose the wallets tab and selected same wallet, but there's no account to selected, which makes the destination account dropdown empty (happens on master too).

let's patch that here

Screenshot 2024-07-29 at 9 57 54 PM

JustinBeBoy:
> I have a wallet with just one account, that is the sending wallet, and i had selected the one account as the source account
> 
> i chose the wallets tab and selected same wallet, but there's no account to selected, which makes the destination account dropdown empty (happens on master too).
> 
> let's patch that here
> 
> Screenshot 2024-07-29 at 9 57 54 PM

So, what should we do?
I think may be we only show wallet tab when application have more than one wallet or one wallet with 2 accounts

JustinBeBoy:
> I have a wallet with just one account, that is the sending wallet, and i had selected the one account as the source account
> 
> i chose the wallets tab and selected same wallet, but there's no account to selected, which makes the destination account dropdown empty (happens on master too).
> 
> let's patch that here
> 
> Screenshot 2024-07-29 at 9 57 54 PM

move to #606

libwallet/assets/btc/rescan.go
```diff
@@ -112,55 +112,6 @@ func (asset *Asset) CancelRescan() {
...
-// and transaction scanning) by stopping the btcwallet, dropping the transaction
-// history from the wallet db, resetting the synced-to height of the wallet
-// manager, restarting the wallet and its chain client, and finally commanding
-// the wallet to resynchronize, which starts asynchronous wallet recovery.
-// Progress of the rescan should be monitored with syncStatus. During the rescan
-// wallet balances and known transactions may not be reported accurately or
-// located. The SPVService is not stopped, so most spvWallet methods will
-// continue to work without error, but methods using the btcWallet will likely
-// return incorrect results or errors.
-func (asset *Asset) rescanAsync() error {
```
dreacot, line 125: why was this file deleted?

ui/cryptomaterial/editor.go
```diff
@@ -83,9 +83,10 @@ type Editor struct {
 
 	isFirstFocus bool
 
-	submitted bool
-	changed   bool
-	selected  bool
+	submitted      bool
+	changed        bool
+	selected       bool
+	isFocusShowHit bool
```
dreacot, line 89:
```suggestion
	showHintOnFocus bool
```

ui/cryptomaterial/editor.go
```diff
@@ -224,6 +225,10 @@ func (e *Editor) Selected() bool {
 	return selected
 }
 
+func (e *Editor) AlwayShowHit() {
```
dreacot, line 228:
```suggestion
func (e *Editor) AlwaysShowHint() {
```

dreacot (changes requested):

libwallet/assets/btc/rescan.go
JustinBeBoy: it not use anywhere so i clean it

dreacot (approved):

Issue #618, opened 2024-08-07
In the Send page, the Amount is not updated after sending


bugjt:
![image](https://github.com/user-attachments/assets/82a86346-d02a-4163-a776-480fb7fbfc9a)
- Error in amount input display unclear.
- Sending from one wallet to another, using max button, after transaction is sent, funds still remain in original wallet.
- noticing cursor position selecting in the password input field is also glitchy

PR #619, opened 2024-08-08
fix the Amount is not updated after sending

bugjt:
Closed: #618 

This PR changed: 
- fix error in amount input display unclear
- disable send button when amount is less than or equal to zero
- update spenable balance after sending and after transaction is synchronized
- refactor wallet/account dropdown
- check for nil values ​​in several places

### DIFF
--- a/ui/page/components/account_dropdown.go
+++ b/ui/page/components/account_dropdown.go
@@ -34,20 +34,27 @@ func NewAccountDropdown(l *load.Load) *AccountDropdown {
 	return d
 }
 
-func (d *AccountDropdown) Setup(wallet sharedW.Asset) *AccountDropdown {
-	if wallet == nil {
+func (d *AccountDropdown) Setup(w sharedW.Asset, args ...*sharedW.Account) *AccountDropdown {
+	if w == nil {
 		return d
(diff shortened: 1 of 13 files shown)


dreacot (commented):
when i use the send max feature, i can't get the input cursor to come to the front. 

i also can't drag to highlight the text in the input field

also i find it weird that the cursor can't be position to any text you want, if i have 4 letter abcd, i should be able to position the cursor after b if i wish to (seems like i have to double tap to achieve that)

Screenshot 2024-08-08 at 4 55 31 PM

when i put an invalid passphrase, the loader icon freezes for about 15 seconds before showing the error message, same freeze also occurs for the right passphrase, not just as long maybe 3 seconds
Screenshot 2024-08-08 at 4 58 11 PM

dreacot (commented):
much better but not butter smooth

https://github.com/user-attachments/assets/1a3cd0a1-9b3d-4084-8e8d-86ddac13644b

bugjt:
It's ready for review

dreacot (approved):

dreacot (approved):

PR #622, opened 2024-08-09
refactor dropdown layout

bugjt:
This PR following changes:
- Refactor dropdown layout
- The dropdown icon in the wallet details page appears to be text so far

image
image

### DIFF
--- a/ui/cryptomaterial/dropdown.go
+++ b/ui/cryptomaterial/dropdown.go
@@ -316,50 +316,14 @@ func (d *DropDown) collapsedAndExpandedLayout(gtx C) D {
 // expandedLayout computes dropdown layout when dropdown is opened.
 func (d *DropDown) expandedLayout(gtx C) D {
 	m := op.Record(gtx.Ops)
-	gtx.Constraints.Max.Y = inf
-	d.layout(gtx, d.ExpandedLayoutInset, d.listItemLayout)
+	gtx.Constraints.Min.Y = gtx.Constraints.Max.Y
+	d.updateDropdownWidth(gtx, true)
+	d.updateDropdownBackground(true)
+	d.ExpandedLayoutInset.Layout(gtx, func(gtx C) D {
+		return d.linearLayout.Layout2(gtx, d.listItemLayout)
+	})
 	op.Defer(gtx.Ops, m.Stop())
 	return D{}
-}
-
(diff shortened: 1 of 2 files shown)


JustinBeBoy:
The width of the dropdown should be equal to the width of the displayed options. I don't think we need to increase width?

image

dreacot:
> The width of the dropdown should be equal to the width of the displayed options

i think you have a point probably should use width of the largest item? but what happens when the largest item is too long, obviously to truncate right?, so what's too long in this case?

JustinBeBoy:
> > The width of the dropdown should be equal to the width of the displayed options
> 
> i think you have a point probably should use width of the largest item? but what happens when the largest item is too long, obviously to truncate right?, so what's too long in this case?

the width of dropdown have define max width, so we don't need do it, if dropdown and dropdown menu have different width so ui not good for me

dreacot:
@bugjt justindo has a good point

let's just use a sane max width and truncate anything longer than that

bugjt:
This means we will have the dropdown and the dropdown menu the same size and both will be the width of the largest item. Is that correct?

bugjt:
We cannot rename names longer than 20 characters and when the name is too long it will be truncated with "..."

image
image

dreacot:
so the truncation is where i was getting to, the screenshot by the left works well. seem it truncates after 10 characters

bugjt:
So we will cut after 10 characters or 20 characters (default maxDropdownItemTextLen = 20)?

bugjt:
If the width of the dropdown is equal to the length of the largest element, it will look like this:
image

dreacot:
image

This is okay

bugjt:
It's ready for review.
image

JustinBeBoy (changes requested):
How can I scroll to see the last wallet?

image

dreacot:
i want to move the dropdown and balance to the same line on mobile.

For the sake of the mobile implementation, i'' say we revert the dropdown to a dynamic width

bugjt:
> i want to move the dropdown and balance to the same line on mobile.
> 
> For the sake of the mobile implementation, i'' say we revert the dropdown to a dynamic width 

I think we should hide "watch only" label on watch only wallet

dreacot:
> I think we should hide "Watch-Only" label on watch only wallet

I agree

bugjt:
It's ready for review.

dreacot (approved):

Issue #621, opened 2024-08-09
app freeze after turn off sync then click back button


JustinBeBoy:
app freeze after turn off sync then click back button

![image](https://github.com/user-attachments/assets/9fae536d-4b21-4cee-b4ef-856b515a4601)

PR #624, opened 2024-08-18
fix app freeze after turn off sync then click back button

bugjt:
closed: #621 

this PR following changes:
- Fix app freeze after turn off sync then click back button. Remove unnecessary RWMutex.Rlock().
- Always return false in the Changed function to prevent the user change the switch when the switch is disabled


dreacot (approved):

JustinBeBoy (approved):
LGTM

Issue #634, opened 2024-09-08
Mobile: Back arrow button / x button too small, Page don't refresh after performing


bugjt:
- Back arrow button too small on all page
- Back arrow button is not aligned with the text
- The segmented control tab is not fitting on the screen properly, I can’t see “Treasury spends well”
- Restore wallet page: The x icon is too small,  missing scroll bar
- Some pages don't refresh after performing an action until I tap somewhere on the screen: Backup modal, Backup page, Keep in mind page, Verify seed words,...

![image](https://github.com/user-attachments/assets/4a29c14c-0c44-430c-8c00-1a70565fb7fb)
![image](https://github.com/user-attachments/assets/faca5b9c-6d6c-4764-b53f-e66508e9900e)
![image](https://github.com/user-attachments/assets/537bd046-20ea-4f65-8d95-713befded5a9)
![image](https://github.com/user-attachments/assets/47193eda-375f-4e24-a777-6afcdeb7132d)
![image](https://github.com/user-attachments/assets/b306f407-1abb-4528-bbfa-7c8da67848c9)
![image](https://github.com/user-attachments/assets/ecf62a3a-2842-413f-88f7-1a349e5801bd)

PR #635, opened 2024-09-09
fix icon button too small on mobile, segment control tab not fit

bugjt:
closed: #634 

This PR has changed:
- Increase icon button size on mobile
- Fix The segmented control tab is not fitting on the governance page
- Add scroll bar to Restore wallet page
- Fix some pages don't refresh after performing an action until tap somewhere on the screen

Screenshot_1725855062
Screenshot_1725855075
Screenshot_1725855125

### DIFF
--- a/ui/page/components/restore_page.go
+++ b/ui/page/components/restore_page.go
@@ -234,9 +234,11 @@ func (pg *Restore) HandleUserInteractions(gtx C) {
 	if pg.tabs.Changed() {
 		pg.tabIndex = pg.tabs.SelectedIndex()
 	}
-
-	if !pg.toggleSeedInput.IsChecked() && pg.toggleSeedInput.Changed(gtx) {
-		pg.seedRestorePage.setEditorFocus()
+	if pg.toggleSeedInput.Changed(gtx) {
+		if !pg.toggleSeedInput.IsChecked() {
+			pg.seedRestorePage.setEditorFocus()
+			pg.ParentWindow().Reload()
+		}
 	}
 
 	if pg.tabIndex == 0 {
--- a/ui/page/seedbackup/backup_instructions.go
+++ b/ui/page/seedbackup/backup_instructions.go
@@ -120,6 +120,11 @@ func (pg *BackupInstructionsPage) OnNavigatedFrom() {}
 // to be eventually drawn on screen.
 // Part of the load.Page interface.
 func (pg *BackupInstructionsPage) Layout(gtx C) D {
+	for i := range pg.checkBoxes {
+		if pg.checkBoxes[i].CheckBox.Update(gtx) {
+			pg.ParentWindow().Reload()
+		}
+	}
 	sp := components.SubPage{
 		Load:       pg.Load,
 		Title:      values.String(values.StrKeepInMind),
--- a/ui/page/seedbackup/save_seed.go
+++ b/ui/page/seedbackup/save_seed.go
@@ -126,6 +126,7 @@ func (pg *SaveSeedPage) OnNavigatedTo() {
 			seed, err := pg.wallet.DecryptSeed(password)
 			if err != nil {
 				m.SetError(err.Error())
+				m.ParentWindow().Reload()
 				return false
 			}
 			m.Dismiss()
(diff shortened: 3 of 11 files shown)


ui/page/components/restore_page.go
```diff
@@ -234,9 +234,11 @@ func (pg *Restore) HandleUserInteractions(gtx C) {
...
 		pg.tabIndex = pg.tabs.SelectedIndex()
 	}
-
-	if !pg.toggleSeedInput.IsChecked() && pg.toggleSeedInput.Changed(gtx) {
-		pg.seedRestorePage.setEditorFocus()
+	if pg.toggleSeedInput.Changed(gtx) {
+		if !pg.toggleSeedInput.IsChecked() {
+			pg.seedRestorePage.setEditorFocus()
+			pg.ParentWindow().Reload()
+		}
```
JustinBeBoy, line 241: why we need break to 2 `if` conditions?

ui/page/seedbackup/backup_instructions.go
```diff
@@ -120,6 +120,11 @@ func (pg *BackupInstructionsPage) OnNavigatedFrom() {}
 // to be eventually drawn on screen.
 // Part of the load.Page interface.
 func (pg *BackupInstructionsPage) Layout(gtx C) D {
+	for i := range pg.checkBoxes {
+		if pg.checkBoxes[i].CheckBox.Update(gtx) {
+			pg.ParentWindow().Reload()
+		}
+	}
```
JustinBeBoy, line 127: I don't think this is best solution, we only need to enable/disable `View seed phrase` button when all boxes checked

ui/page/seedbackup/save_seed.go
```diff
@@ -126,6 +126,7 @@ func (pg *SaveSeedPage) OnNavigatedTo() {
 			seed, err := pg.wallet.DecryptSeed(password)
 			if err != nil {
 				m.SetError(err.Error())
+				m.ParentWindow().Reload()
```
JustinBeBoy, line 129: why we need reload in here? I see we don't have reason to do it

ui/page/seedbackup/verify_seed.go
```diff
@@ -195,6 +195,7 @@ func (pg *VerifySeedPage) verifySeed() {
 				}
 
 				m.SetError(err.Error())
+				m.ParentWindow().Reload()
```
JustinBeBoy, line 198: it the same above

ui/page/wallet/single_wallet_main_page.go
```diff
@@ -328,6 +328,10 @@ func (swmp *SingleWalletMasterPage) changeTab(tab string) {
 // displayed.
 // Part of the load.Page interface.
 func (swmp *SingleWalletMasterPage) HandleUserInteractions(gtx C) {
+	if swmp.checkBox.CheckBox.Update(gtx) {
+		swmp.ParentWindow().Reload()
```
JustinBeBoy, line 332: why we need reload in here? I see we don't have reason to do it

JustinBeBoy (changes requested):
- Bottom margin too small

image

- loading not work when click confirm
image

ui/page/seedbackup/save_seed.go
bugjt: It was added to fix the modal not refreshing after performing an action until somewhere on the screen is touched

ui/page/seedbackup/verify_seed.go
bugjt: same above

ui/page/wallet/single_wallet_main_page.go
bugjt: same above

ui/page/seedbackup/backup_instructions.go
mpt-tiendc: It was added to fix the button not active after I check all checkbox until somewhere on the screen is touched

bugjt:
> * Bottom margin too small
> 
> image
> * loading not work when click confirm
> 
> image

ok I will add more margin to bottom of button

Issue #665, opened 2024-09-20
verify a wallet name is available before allowing users to input their seed


dreacot:
<img src=https://github.com/user-attachments/assets/c7e28cde-4800-4f1a-97b6-d61ea1d306cc height=500 />

dreacot:
resolved by https://github.com/crypto-power/cryptopower/pull/669

Issue #652, opened 2024-09-20
prevent body of the wallet page from sliding as it causes users to accidentally scroll into another tab


dreacot:
swiping left and right takes you to the next or previous tab, but i find it to accidentally do that sometimes

<img src=https://github.com/user-attachments/assets/c290da90-837c-40b2-9fd2-1c7eeac795da height=500 />

dreacot:
resolved by https://github.com/crypto-power/cryptopower/pull/669

Issue #649, opened 2024-09-20
clicking backup wallet from inside the wallet page should not show the top balance bar


dreacot:
<img src=https://github.com/user-attachments/assets/345f032e-06a2-4ad7-a7c1-4d69dfb8149f height=500 />
<img src=https://github.com/user-attachments/assets/859b44ef-637a-4b38-9c46-a5e5d3212983 height=500 />

dreacot:
resolved by https://github.com/crypto-power/cryptopower/pull/669

Issue #648, opened 2024-09-20
creating new wallet from the exchange page doesn't show full right margins


dreacot:
- [x] creating new wallet from the exchange page doesn't show full right margins
- [x] fix the top margins
- [x] we probably shouldn't show the usd top bar and segmented control when the create wallet page is showing

<img src=https://github.com/user-attachments/assets/de31f918-4ed7-4f3b-a683-1d3295a93538 height=500 />
<img src=https://github.com/user-attachments/assets/97af8dab-5b4a-42f9-981a-ae57d73bdf79 height=500 />

dreacot:
resolved by https://github.com/crypto-power/cryptopower/pull/669

PR #669, opened 2024-09-23
show fullscreen when create/backup wallet, verify a wallet name before continue

bugjt:
fixed:
- #649 
- #648 
- #652 
- #665 

This PR following changed:
- Show fullscreen when backup wallet
- Show fullscreen when create wallet
- Disable swipe on body of the wallet page
- Verify a wallet name is available before allowing users to input their seed


libwallet/assets_manager.go
```diff
@@ -615,6 +616,18 @@ func (mgr *AssetsManager) DeleteBadWallet(walletID int) error {
 	return nil
 }
 
+// Check if wallet name is already exists
```
dreacot, line 619:
```suggestion
// Check if wallet name already exists
```

JustinBeBoy (changes requested):
1. need margin title the same setting page
r1
r2

2. remove topbar in mobile and fix margin right
r4

3. Can't switch tab on mobile
r5

4. label's error not disappear after typing 
r6

bugjt:
@JustinBeBoy fixed 1,2,3,4. 
2. I think should not hide topbar

dreacot (approved):

JustinBeBoy (approved):
LGTM

Issue #690, opened 2024-10-18
app crashed while listening for Orderbook Notifications


dreacot:
i just had the app running in the background and came back to this crash

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x28 pc=0x105cfb3c8]

goroutine 2373 [running]:
github.com/crypto-power/cryptopower/ui/page/dcrdex.(*DEXMarketPage).listenForOrderbookNotifications(0x14006066000)
        /Users/user/go/src/github.com/crypto-power/cryptopower/ui/page/dcrdex/market.go:454 +0x78
github.com/crypto-power/cryptopower/ui/page/dcrdex.(*DEXMarketPage).fetchOrderBook.func1()
        /Users/user/go/src/github.com/crypto-power/cryptopower/ui/page/dcrdex/market.go:432 +0x174
created by github.com/crypto-power/cryptopower/ui/page/dcrdex.(*DEXMarketPage).fetchOrderBook in goroutine 1798
        /Users/user/go/src/github.com/crypto-power/cryptopower/ui/page/dcrdex/market.go:423 +0x2b8
exit status 2
```

dreacot:
resolved by https://github.com/crypto-power/cryptopower/pull/692

PR #692, opened 2024-10-19
Cleanup dex market page

ukane-philemon:
- Don't read feed channel from pg.selectedMarketOrderbook
- Remove unused code
- Calculate dex order by lot
- Display order Amount: this will make it less confusing for user when calculating what they get with amount of lots they want.

Screenshot is stale: `Quantity -> Amount`

Screenshot 2024-10-19 at 1 42 07 AM


norwnd:
Hey @ukane-philemon I've briefly took a look at https://github.com/crypto-power/cryptopower/issues/690 and noticed there is quite a bit of concurrent interaction happening with `DEXMarketPage struct` but not enough syncronization between go-routines involved,

I don't plan to investigate those myself, just pointing out this is probably needs addressing and that it can cause https://github.com/crypto-power/cryptopower/issues/690 and similar issues.

ui/page/dcrdex/market.go
```diff
@@ -451,7 +448,7 @@ func (pg *DEXMarketPage) listenForOrderbookNotifications() {
 		select {
 		case <-pg.ctx.Done():
 			return
-		case bookUpdate, ok := <-pg.selectedMarketOrderBook.feed.Next():
```
ukane-philemon, line 454: RE: https://github.com/crypto-power/cryptopower/pull/692#issuecomment-2423800125

Yes, this caused the panic since we run `listenForOrderbookNotifications` in a goroutine and still try to access `pg.selectedMarketOrderBook.feed` which might become nil when the page state changes. 

I think this is just the only case where this might happen on this page, but will look again.

ukane-philemon: > but will look again

Yes, that's the only case where we might have an issue.

dreacot (approved):

PR #699, opened 2024-10-30
update neutrino version

dreacot:
This is expected to fix a block cannot be found sync error for btc spv.

https://github.com/decred/dcrdex/pull/3038


ukane-philemon (approved):
LGTM

PR #719, opened 2025-01-30
improve dex screens on mobile

dreacot:


ukane-philemon:
Screenshot 2025-01-30 at 2 50 40 PM

I think we could introduce short forms for mobile to prevent ugly breaks like this, e.g 24h Vol.(DCR)?

ui/page/root/home_page.go
```diff
@@ -138,15 +138,15 @@ func (hp *HomePage) initPageItems() {
...
-		navigationTabTitles = append(navigationTabTitles[:insertionPoint], append([]string{values.String(values.StrTrade)}, navigationTabTitles[insertionPoint:]...)...)
+	// if !appos.Current().IsDarwin() {
+	// Determine the insertion point, which is second to last position
+	insertionPoint := len(navigationTabTitles) - 1
+	if insertionPoint < 0 {
+		insertionPoint = 0
 	}
+	// Append at the second to last position
+	navigationTabTitles = append(navigationTabTitles[:insertionPoint], append([]string{values.String(values.StrTrade)}, navigationTabTitles[insertionPoint:]...)...)
+	// }
```
ukane-philemon, line 149: oops?

ukane-philemon (changes requested):

ui/page/root/home_page.go
dreacot: good catch

ukane-philemon:
> > Screenshot 2025-01-30 at 2 50 40 PM
> > I think we could introduce short forms for mobile to prevent ugly breaks like this, e.g 24h Vol.(DCR)?
> 
> Screenshot 2025-01-30 at 3 36 43 PM looks like each element moves independently

Ah I see.

PR #720, opened 2025-01-30
Allow `SegmentedControl` translate titles on the fly

ukane-philemon:
This PR changes the required segmented titles from pre-translated to the translation key to allow SegmntedControl to translate on the fly and be able to handle cases where language change might result to incorrect segmented title equality checks.

`SegmentedControl.SetSelectedSegment` now accepts an actual translation key instead of a pre-translated string and `SegmentedContol.SelectedSegment` returns the same.

### DIFF
--- a/ui/page/send/send_destination.go
+++ b/ui/page/send/send_destination.go
@@ -15,8 +15,8 @@ import (
 )
 
 var tabOptions = []string{
-	values.String(values.StrAddress),
-	values.String(values.StrWallets),
+	values.StrAddress,
+	values.StrWallets,
 }
 
 type destination struct {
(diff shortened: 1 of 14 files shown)


ui/page/send/send_destination.go
```diff
@@ -167,7 +167,7 @@ func (dst *destination) clearAddressInput() {
 // isSendToAddress returns the current tab selection status without depending
 // on a buffered state.
 func (dst *destination) isSendToAddress() bool {
-	return dst.accountSwitch.SelectedSegment() == values.String(values.StrAddress)
+	return dst.accountSwitch.SelectedSegment() == values.StrAddress
```
ukane-philemon, line 170: This is one case where the formal approach caused the send page to not function properly due to an incorrect equality check.

`SelectedSegment()` was returning a pre-translated string(english) and  `values.String(values.StrAddress)` was returning the current string translation(e.g french) after a language change.

dreacot (approved):

Issue #644, opened 2024-09-20
while on the tx history page, the incoming blocks/confirmations are not registered


dreacot:
it seems like leaving the page, then returning then updates the confirmations

PR #728, opened 2025-02-25
Update total balance and total USD balance on Overview screen

vufon:
Related issue: #643 

https://github.com/user-attachments/assets/8a6051ad-c722-454f-8b30-2c4338e96524

PR #729, opened 2025-02-25
Fix the bug where the transaction history was not updated when a new transaction or confirmation arrived

vufon:
Related issue: #644 

https://github.com/user-attachments/assets/f5410c66-d501-47f0-b788-88ade36ad480


dreacot:
Based on my comment https://github.com/crypto-power/cryptopower/pull/728#issuecomment-2684287322 this page is also under home_page.go so i believe a global check would reduce some duplicate.

What do you think?

vufon:
> Dựa trên bình luận [#728 của tôi (bình luận),](https://github.com/crypto-power/cryptopower/pull/728#issuecomment-2684287322) trang này cũng nằm trong home_page.go nên tôi tin rằng việc kiểm tra toàn cục sẽ giảm bớt một số nội dung trùng lặp.
> 
> Bạn nghĩ sao?

Yes sir, I got it

PR #733, opened 2025-03-03
Update balance and tx history when new tx is registered. Clean up listener structure between parent and subpage

### LINKED ISSUE #643: the total usd balance doesn't properly update, when the balance of one of the wallet changes
- [ ] the total usd balance doesn't properly update, when the balance of one of the wallet changes, it only updates when going to settings page and returning

- [ ] often times the selected exchange rate resets to none

### LINKED ISSUE #644: while on the tx history page, the incoming blocks/confirmations are not registered

vufon:
Resolves #643  Resolves #644

- Cleaned up to handle new tx listener on homepage (or single wallet main page with wallet details)
- Updated on pages: Overview, Transactions, Wallet info page, Wallet detail transactions


ui/page/transaction/transactions_page.go
```diff
@@ -757,28 +756,12 @@ func exportTxs(assets []sharedW.Asset, fileName string) error {
...
-	for _, w := range pg.assetWallets {
-		w.RemoveTxAndBlockNotificationListener(TransactionsPageID)
-		err := w.AddTxAndBlockNotificationListener(txAndBlockNotificationListener, TransactionsPageID)
-		if err != nil {
-			log.Errorf("Error adding tx and block notification listener: %v", err)
-			return
-		}
+// Update transaction list when there is new tx or new confirmed status
+func (pg *TransactionsPage) ListenForTxNotification(walletID int) {
+	if pg.selectedWallet != nil && pg.selectedWallet.GetWalletID() != walletID {
```
dreacot, line 761: what happens if the selected wallet is nil? because you have the option to view transactions across all wallets

looks like this expects a wallet to always be selected which is not always the case

vufon: I think when selected wallet is nil, this logic is 'false' and it will continue to the next statement:
pg.selectedWallet != nil: false
so it will cover the cases

dreacot (approved):

Issue #730, opened 2025-02-26
Process notifications whenever new transactions arrive


vufon:
Currently, the notification is displayed only when the user is on the wallet details page.
When the user is on external screens or the settings page, no notification is shown
I see that the new transaction notification should be displayed regardless of which screen the user is on. The notification content may need further adjustments

Issue #727, opened 2025-02-22
Can't set Transaction Notification to enable in settings


vufon:
Cannot enable Transaction Notification.
I think it may be due to the status check location of this Switch. I am not sure why transaction notification status is only enabled when privacy is on?
`func (mgr *AssetsManager) IsTransactionNotificationsOn() bool {
	var data bool
	mgr.ReadAppConfigValue(sharedW.TransactionNotificationConfigKey, &data)
	return data && mgr.IsPrivacyModeOn()
}`

I think the logic in this area needs to be clarified. Or if there is a reason why this Switch cannot be turned on, there should be some kind of notification for the user to understand the reason.
https://github.com/user-attachments/assets/9b179ed6-288e-41a0-a0b9-06c07433452f
https://github.com/user-attachments/assets/df9d2a33-b751-46b3-8f74-b1aa99912610

Issue #650, opened 2024-09-20
transaction notification does not send push notifications on mobile


dreacot:
- [ ] transaction notification does not send push notifications on mobile
- [ ] Clicking on notification sent on desktop don't take you back into the app

<img src=https://github.com/user-attachments/assets/b4cb830c-686f-437e-a4a6-325cd52a7414 height=500 />

vufon:
I'll research and implement this issue

PR #734, opened 2025-03-03
Improve push notification

vufon:
Resoved issue: #650 #727 #730


dreacot:
this looks like it has been replaced by https://github.com/crypto-power/cryptopower/pull/733

PR #744, opened 2025-08-24
Added a media query fix for when the device height is less than 600.

### LINKED ISSUE #742: stop at this step, it will not work!
//[

Image

](url)

Josephus-git:
closes #742 
fixed rendering issue on create wallet button for small devices 
Screenshot from 2025-08-24 22-37-51

### DIFF
--- a/ui/page/start_page.go
+++ b/ui/page/start_page.go
@@ -398,6 +398,10 @@ func (sp *startPage) welcomePage(gtx C) D {
 
 func (sp *startPage) loadingSection(gtx C) D {
 	return sp.pageLayout(gtx, func(gtx C) D {
+
+		// Media query: check if device height is less than 600
+		isSmallScreen := gtx.Constraints.Max.Y < gtx.Dp(600)
+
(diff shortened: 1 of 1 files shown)


ui/page/start_page.go
```diff
@@ -450,8 +454,16 @@ func (sp *startPage) loadingSection(gtx C) D {
...
-				inset := layout.Inset{Top: values.MarginPadding100}
-				if sp.IsMobileView() {
+
+				var topMargin unit.Dp
+				if isSmallScreen {
+					topMargin = values.MarginPadding10
+				} else {
+					topMargin = values.MarginPadding100
+				}
+				inset := layout.Inset{Top: topMargin}
```
dreacot, line 464:
```suggestion
				topMargin := values.MarginPadding100
				if isSmallScreen {
					topMargin = values.MarginPadding10
				}
				inset := layout.Inset{Top: topMargin}
```

Try to avoid if else statements where possible as it makes the code messy. the above is shorter and easier to read

dreacot (changes requested):

